PR #6, opened 2021-03-19
define feed component so metamask instructions are in their correct p…

claireolmstead:
# Problem

As a user who has not yet created a wallet using metamask, I would like to be prompted with instructions on how to do that.
[link to Pivotal Tracker #177346917](https://www.pivotaltracker.com/story/show/177346917)

# Solution

Create `<MetamaskInstructions />` component that is shown in the feed block given the condition that the user has not yet downloaded metamask.

## Change summary:

- Create `<MetamaskInstructions />` component with instruction text and links
- Create `<Feed />` component, so the instructions live in the correct place
- Delete some of the boiler plate components and styling
- TODO: Add condition of of when to show instructions and when not to show instructions
- TODO: add styling

## Steps to Verify:

1. uninstall metamask
2. follow instructions to check that they are clear
3. check that on future logins this component does not display

## Screenshots (optional):

Show-n-tell images/animations here


claireolmstead:
Decided to not use Metamask, but to embed the Torus wallet.  In favor of that decision, we are going to close out this PR.

PR #7, opened 2021-03-19
Feature/features set/177209775

claireolmstead:
# Problem

As a developer, I want to know the scope of the Example Client, so I know what the end goal is or what I need to work on.
[link to Pivotal Tracker #177209775](https://www.pivotaltracker.com/story/show/177209775)

# Solution

Add a section in the README.md that describes a list of features that will be included in the Example Client.

## Change summary:

- Going off of the stories in Pivotal of what we are building for our V1, I created a list of main features of the EC
- Included a note at the bottom that the list of features is subject to change

## Steps to Verify:

1. Check that the list is correct
2. Check that the list is in a format/organized how we want it to be

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,33 @@ __Testing:__
 * Tests written in [Jest](https://jestjs.io) and [Enzyme](https://enzymejs.github.io/enzyme/)
 * CI/CD done with [Github Actions](https://github.com/features/actions)
 
+***
+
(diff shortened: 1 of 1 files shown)


wilwade (changes requested):
Looks like a lot of extra files got in here. However the README section looks good!

wilwade (approved):

PR #8, opened 2021-03-19
setup basic component structure and formatting

claireolmstead:
# Problem

As a developer, I want to have the component structure roughly setup, along with basic formatting, so that when we build out the first features, there is a sense of organization.
[link to Pivotal Tracker #12345678](https://www.pivotaltracker.com/story/show/12345678)

# Solution
- Build the general component structure
- Add formatting only
 - No brand-influenced styling
- Add tests to make sure components render

## Change summary:

- Created the following empty components
 - Header
 - Feed
 - NewPost
 - PostList
 - Profile
- Created tests to make sure they render
- Added styling to for layout and to show layout

## Steps to Verify:

1. Run EC and see that everything is rendering correctly
2. Check that the component structure is correct

## Screenshots (optional):
Screen Shot 2021-03-19 at 4 17 10 PM


src/App.scss
```diff
@@ -1,43 +1,10 @@
...
-  height: 40vmin;
-  pointer-events: none;
-}
-
-.App__header {
-  @include headline-text;
-
-  background-color: $gray-28;
-  min-height: 100vh;
+.Content__block {
```
claireolmstead, line 7: Is it weird to put this styling here? It matches the file format, but feels a little weird under the imports.  What's the right call?

.gitignore
```diff
@@ -24,3 +24,5 @@ yarn-error.log*
 
 .idea
 junit.xml
+src/App.css
+src/App.css.map
```
wilwade, line 28: If we git rm these files, we do not need to ignore them

src/App.scss
wilwade: I'd likely name it `.App__content` instead, that way it follows more of BEM, but having this file as the App component's styles is fine by me.

wilwade (changes requested):
Two small notes.

wilwade (approved):

PR #9, opened 2021-03-26
Feature/login with wallet 177347812

claireolmstead:
# Problem
- As a user, I obtain a social identity in the form of a smart contract by creating  an account or logging in with Torus.
- As a user, I am able to create multiple social identities.

[link to Pivotal Tracker #177515975](https://www.pivotaltracker.com/story/show/177515975)

# Solution
- Look to Torus integration in friendly
- Basically copy over what we need

with @claireolmstead 

## Change summary:
- create fake sdk file
- add torus config files/setup
- add `onAuthenticate()` and `logout()` functionality
- add necessary `/services` and `/utilities` files
- rearrange file structure for scss and test files into their own directories
- add styling to login/logout button
- add `antd-overrides.scss`
- add `.stylelintignore` file
- get tests working

## Steps to Verify:

1. Make sure it works for the following cases:
- case 1: no Torus connected
- connect the wallet

case 2: I do not have a dsnp identity
- create a dsnp identity
- sdk call

case 3: I already have a dsnp identity
- I can just login with that identity
- sdk call

## Screenshots (optional):
Screen Shot 2021-03-26 at 4 09 10 PM
Screen Shot 2021-03-26 at 4 09 29 PM


.gitignore
```diff
@@ -23,4 +23,5 @@ yarn-debug.log*
 yarn-error.log*
 
 .idea
-junit.xml
```
claireolmstead, line 26: test was not passing without this there.

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
```
claireolmstead, line 6: delete

.idea/workspace.xml
```diff
@@ -0,0 +1,212 @@
...
+        <test-name value="testEnabler" />
+      </test-names>
+      <method v="2" />
+    </configuration>
+    <configuration name="start" type="js.build_tools.npm">
+      <package-json value="$PROJECT_DIR$/package.json" />
+      <command value="start" />
+      <node-interpreter value="project" />
+      <envs />
+      <method v="2" />
```
claireolmstead, line 142: delete

src/components/Login.tsx
```diff
@@ -0,0 +1,98 @@
...
+      console.log("connAddr: ", walletAddress);
+      const socialAddress = await sdk.getSocialIdentity(walletAddress);
+      const profile = await sdk.getProfile(socialAddress);
+      const graph = await sdk.getGraph(socialAddress);
+      auth(walletAddress, socialAddress, profile as Profile, graph as Graph);
+      startLoading(false);
+    } catch (error) {
+      if (error.code === 4001) {
+        // EIP 1193 userRejectedRequest error
+        setAlertError("Please connect to MetaMask.");
```
claireolmstead, line 54: Figure out what the error code actually is and display correct alert.

src/components/Login.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <div className="Login__block">
+      {alertError && (
+        <Alert
+          type="error"
+          message={alertError}
+          banner
+          closable={true}
+          onClose={() => setAlertError("")}
+        />
+      )}
```
claireolmstead, line 75: style alert

src/components/test/Login.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+  it("renders without crashing", () => {
+    expect(() => {
+      shallow(
+        <Login
+          logout={jest.fn}
+          onAuthenticate={jest.fn}
+          socialAddress={mockSocialAddress}
+        />
+      );
+    }).not.toThrow();
```
claireolmstead, line 18: do we want to test anything else or is it too much of a pain in the butt for an error that would be very obvious

src/services/mocksdk.ts
```diff
@@ -0,0 +1,27 @@
+import { generateRandomSocialGraph } from "../test/testGraphs";
+import { generateProfile } from "../test/testProfiles";
+import { Graph, HexString, Profile } from "../utilities/types";
+
```
claireolmstead, line 4: check out this story to see what I want to do.  what blockers do we have on implementing this?

src/services/wallets/torus.js
```diff
@@ -0,0 +1,239 @@
+import web3Torus from "./tweb3";
+
+export const getAccount = async () => {
+  return await web3Torus.torus.accounts;
+  //web3Torus.web3.eth.getAccounts().then(accounts => {
+  //  return accounts[accountIndex];
+  //});
```
claireolmstead, line 7: can we delete comments?

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
claireolmstead, line 1: file should be deleted and .idea/* ignored in gitignore

.idea/workspace.xml
```diff
@@ -0,0 +1,212 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
claireolmstead, line 1: delete file and add to gitgnore

src/App.tsx
```diff
@@ -1,19 +1,92 @@
...
+  clearSession,
+  getSession,
+  hasSession,
+  setSession,
+  updateSession,
+} from "./services/session";
 import Header from "./components/Header";
 import Feed from "./components/Feed";
-import Profile from "./components/Profile";
+import ProfileBlock from "./components/ProfileBlock";
```
claireolmstead, line 14: do we like the naming of `ProfileBlock` or is there something better?

src/components/Login.tsx
```diff
@@ -0,0 +1,97 @@
...
+    try {
+      await torus.enableTorus();
+      const walletAddress = await torus.getWalletAddress();
+      const socialAddress = await sdk.getSocialIdentity(walletAddress);
+      const profile = await sdk.getProfile(socialAddress);
+      const graph = await sdk.getGraph(socialAddress);
+      auth(walletAddress, socialAddress, profile as Profile, graph as Graph);
+      startLoading(false);
+    } catch (error) {
+      setAlertError(error.toString());
```
claireolmstead, line 51: rad! happy you figured that out

src/components/Login.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+    startLoading(false);
+  };
+
+  const torusLogout = () => {
+    logout();
+    if (torus.isInitialized()) {
+      torus.logout();
+    }
```
claireolmstead, line 60: good catch!

src/components/ProfileBlock.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { HexString, Profile } from "../utilities/types";
+
+interface ProfileProps {
+  walletAddress: HexString | null;
+  socialAddress: HexString | null;
+  profile: Profile | null;
```
claireolmstead, line 7: do we want to include the `or null` in the types file instead of having to iterate it throughout?

src/components/test/Login.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    it("button triggers logout sequence", () => {
+      const component = shallow(
+        <Login
+          logout={jest.fn}
+          onAuthenticate={jest.fn}
+          socialAddress={mockSocialAddress}
+        />
+      );
+      component.find("Button").simulate("click");
+      expect(torusLogoutMock).toHaveBeenCalled();
```
claireolmstead, line 74: this is so rad! thank you for writing these tests!

.idea/vcs.xml
Ajaxan: Yup. These files come from some IDEs, we can add more IDE generated as/if they pop up.

src/App.tsx
Ajaxan: I think in general, I'd prefer our types to have the proper names. So I think our components may need a slightly different naming convention. We could go with entirely different names but in general that may get confusing.

So I think for now ProfileBlock works for me. If we really want a cleaner name we could go with Account, but again, the alternate names may get confusing. At least ProfileBlock is pretty obvious on it's function.

src/components/ProfileBlock.tsx
Ajaxan: So like make some proper WalletAddress and SocialAddress types? Yeah sure. I can do that. I'll add HexAddress that can be either a HexString or null.

Ajaxan: I think the right way to do this is instead to simply not pass walletAddress and socialAddress unless they exist and we can make these params optional.

Ajaxan:
Some comments about the latest commit and future plans. Especially based around my sync with @enddynayn .

1. We will be replacing the use of global react state currently in place with redux in this story: https://www.pivotaltracker.com/story/show/177598915
2. We will be modifying the sessionStorage to store our redux state in this story: https://www.pivotaltracker.com/story/show/177602937
3. We will be creating a generalized walletAPI for interfacing with the planned multiple wallet login system in this story: https://www.pivotaltracker.com/story/show/177602979

src/components/Login.tsx
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  return (
+    <div className="Login__block">
+      {alertError && (
+        <Alert
+          className="Loin__alert"
```
shannonwells, line 67:
```suggestion
          className="Login__alert"
```

shannonwells (changes requested):
Someone checked in a swap file, src/components/.Login.tsx.swp -- this should be removed and .swp files added to .gitignore

src/services/sdk.ts
```diff
@@ -0,0 +1,33 @@
...
+): Promise<HexString> => {
+  let socialAddress: HexString = await sdk.getSocialIdentityfromWalletAddress(
+    walletAddress
+  );
+  if (!socialAddress) {
+    socialAddress = await sdk.createSocialIdentityfromWalletAddress(
+      walletAddress
+    );
+  }
+  return socialAddress;
```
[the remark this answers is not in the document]
Ajaxan: The thinking was that it shouldn't ever be null. Of course depending on the way the SDK works this may need to be reworked. We first ask the SDK for the socialAddress, and if that's null we create one. In either case we should get back a social identity unless there is an error.

shannonwells (approved):
I didn't get a chance to run the code, but the code itself looks good.

PR #12, opened 2021-04-05
Feature/build profile component/177348143

claireolmstead:
# Problem

As a user, I want to view my profile information and others' profile information.
[link to Pivotal Tracker #177348143](https://www.pivotaltracker.com/story/show/177348143)

# Solution

follow the wire frames to create the component and styling that displays your profile information

## Change summary:

- created `userAvatar` component to have a reusable avatar component
- created component that houses profile elements
- created component that builds list of following/follows me list
- created temp data structure with relationship info
- added styling

## Steps to Verify:

1. make sure code structure looks good
2. make sure it looks like the wire frame


Ajaxan:
I know its still a draft, but one thing to watch out for. All the current login code is being overwritten and some old files that were removed re-appeared. I'd drop the 3 commits from login and rebase on the main branch. 

Also, you are removing 60k lines which suggest an error with npm versions and the package-lock.json. What version are you using?

claireolmstead:
> I know its still a draft, but one thing to watch out for. All the current login code is being overwritten and some old files that were removed re-appeared. I'd drop the 3 commits from login and rebase on the main branch.
> 
> Also, you are removing 60k lines which suggest an error with npm versions and the package-lock.json. What version are you using?

@Ajaxan there is a ton of cleanup to do.  i was using node v 14.5.0 and it was not compatible after the rebase, which I was thinking was pretty weird.  Planning on looking deeper into it tomorrow (today is a website day) unless you would like to poke around.  Thank you for taking a look!

src/components/ProfileBlock.tsx
```diff
@@ -1,15 +1,100 @@
...
+      }
+    }
+  }, [isEditing, newName, newHandle]);
   return (
     <div className="Profile__block">
-      <h1>Profile</h1>
-      <div>Wallet Address: {profile?.walletAddress}</div>
-      <div>Social Address: {profile?.socialAddress}</div>
-      <div>Profile Name: {profile?.name || "null"}</div>
+      {socialAddress || !socialAddress ? (
```
claireolmstead, line 33: get rid of !socialAddress when login is working again

src/components/ProfileBlock.tsx
```diff
@@ -1,15 +1,100 @@
...
+        <>
+          <div className="Profile__personalInfoBlock">
```
claireolmstead, line 35: change classnames to ProfileBlock

src/components/ProfileBlock.tsx
```diff
@@ -1,15 +1,100 @@
...
+                >
+                  {isEditing ? "SAVE" : "EDIT"}
+                </Button>
+              )}
+            </div>
+            <div className="Profile__personalInfo">
+              <label className="Profile__personalInfoLabel">NAME</label>
+              <input
+                className={
+                  isEditing ? "Profile__name Profile__editing" : "Profile__name"
```
claireolmstead, line 62: make this a function

src/components/ProfileBlock.tsx
```diff
@@ -1,15 +1,100 @@
...
+                }
+                value={newName || newName === "" ? newName : profileName}
+                onChange={(e) => setNewName(e.target.value)}
+                disabled={!isEditing}
+              />
+              <label className="Profile__personalInfoLabel">HANDLE</label>
+              <input
+                className={
+                  isEditing
```
claireolmstead, line 71: call function

src/components/ProfileBlock.tsx
```diff
@@ -1,15 +1,100 @@
...
+                value={newHandle || newHandle === "" ? newHandle : handle}
+                onChange={(e) => setNewHandle(e.target.value)}
+                disabled={!isEditing}
+              />
+              <label className="Profile__personalInfoLabel">
+                SOCIAL ADDRESS
+              </label>
+              <input
+                className={
+                  isEditing
```
claireolmstead, line 84: call function

src/components/UserAvatar.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface UserAvatarProps {
+  socialAddress: HexString | null;
+  profile: Profile | null;
+  avatarSize: string;
+}
+
+const UserAvatar = ({
+  socialAddress,
+  profile,
```
claireolmstead, line 20: only need profile.  Should not get from redux store right?

src/components/scss/Profile.scss
```diff
@@ -1,7 +1,74 @@
...
+}
+
+.Profile__personalInfoLabel {
+  color: $text-gray;
+  line-height: 1;
+}
+
+.Profile__name {
+  @include large-text;
+  font-weight: 550;
```
claireolmstead, line 35: maybe set a variable or mixin as 550 so it can uniform all of the bolded font-weights

claireolmstead:
Just realized I forgot to write tests. those will be coming shortly.

src/components/FollowersFollowing.tsx
```diff
@@ -0,0 +1,86 @@
...
+    if (selectedListTitle === listTitle) {
+      setSelectedListTitle("");
+      return;
+    }
+    if (listTitle === "showFollowers") {
+      setSelectedListTitle(listTitle);
+    } else {
+      setSelectedListTitle(listTitle);
+    }
+  };
```
wilwade, line 43: This is for a toggle right?

```suggestion
  const toggleSelectedList = (listTitle: string) => {
    if (selectedListTitle === listTitle) setSelectedListTitle("");
    else setSelectedListTitle(listTitle);
  };
```

src/components/FollowersFollowing.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { useState } from "react";
+import { Button } from "antd";
+import FollowersFollowingUsers from "./FollowersFollowingUsers";
+import { preFabProfiles } from "../test/testProfiles";
+
+const FollowersFollowing = (): JSX.Element => {
+  const [selectedListTitle, setSelectedListTitle] = useState<string>("");
```
wilwade, line 7: I might switch this to an enum instead of a string. There are only three states:

```typescript
enum ListStatus = {
  CLOSED;
  FOLLOWERS;
  FOLLOWING;
}
```

That way instead of string comparisons, we get type safety and completeness.

src/components/FollowersFollowingUsers.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import UserAvatar from "./UserAvatar";
+import { Button } from "antd";
+
+interface FollowersFollowingUsersProps {
+  selectedListTitle: string;
```
wilwade, line 6: Here to it would be nice to have the enum for safety

src/components/FollowersFollowingUsers.tsx
```diff
@@ -0,0 +1,42 @@
...
+  tempUserList,
+}: FollowersFollowingUsersProps): JSX.Element => {
+  const userList = () => {
+    switch (selectedListTitle) {
+      case "showFollowers":
+        return tempUserList.filter((user) => user.followsMe);
+      case "showFollowing":
+        return tempUserList.filter((user) => user.following);
+      default:
+        return [];
```
wilwade, line 21: Another reason to use the enum is that you don't have to have a default case! (It would instead be used with ListStatus.CLOSED and TS will check for completeness in the switch statement)

src/components/ProfileBlock.tsx
```diff
@@ -1,15 +1,113 @@
...
+                <Button
+                  className="Profile__editButton"
+                  onClick={() => setIsEditing(!isEditing)}
+                >
+                  EDIT
+                </Button>
+              )}
+            </div>
+            <div className="Profile__personalInfo">
+              <label className="Profile__personalInfoLabel">NAME</label>
```
wilwade, line 82: You can use CSS to make these uppercase. That way screen readers will handle it better. (also you don't have to edit each one when later you need to change it to not all uppercase)

src/components/test/UserAvatar.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import { shallow } from "enzyme";
+import UserAvatar from "../UserAvatar";
+import { getPrefabProfile } from "../../test/testProfiles";
+
+describe("UserAvatar", () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
```
wilwade, line 8: I don't think you have any mocks here... yet.

src/index.css
```diff
@@ -0,0 +1,12 @@
+body {
```
wilwade, line 1: Is this also compiled code?

src/index.css.map
```diff
@@ -0,0 +1 @@
+{"version":3,"sourceRoot":"","sources":["index.scss"],"names":[],"mappings":"AAMA;EACE;EACA;;;AAWF;EACE;EACA;EACA","file":"index.css"}
```
wilwade, line 1: compiled code

src/services/wallets/torus.ts
```diff
@@ -47,9 +47,10 @@ export const getPublicAddress = async (
 };
 
 export const getWalletAddress = async (): Promise<string> => {
-  const { verifier, verifierId } = await getUserInfo();
+  //const { verifier, verifierId } = await getUserInfo();
+  const { verifierId } = await getUserInfo();
   const walletAddressHolder = await getPublicAddress(
-    verifier as VerifierTypes,
+    "google" as VerifierTypes,
```
wilwade, line 53: I'm guessing this is waiting for #18 ?

src/theme/antd-overrides.scss
```diff
@@ -1,9 +1,72 @@
...
+  color: $black !important;
+  background: $white !important;
+}
+
+.FollowersFollowing__buttonBlock .ant-btn:first-child {
+  border-radius: 8px 0 0 8px;
+  border-width: 2px 1px 2px 2px;
+}
+
+.Profile__editButton {
```
wilwade, line 61: Looks like a non-override sneaked in

src/theme/antd-overrides.scss
```diff
@@ -1,9 +1,72 @@
...
+  }
+
+  &:hover, &:active {
+    border-color: $black;
+    background: $white;
+    color: $black;
+  }
+}
+
+.FollowersFollowing__button--active {
```
wilwade, line 51: Looks like a non-override sneaked in

wilwade (changes requested):

src/components/FollowersFollowing.tsx
claireolmstead: duhh!! thanks for catching that

src/services/wallets/torus.ts
claireolmstead: Yes.  there is a bug here and this is a work around

src/components/FollowersFollowing.tsx
```diff
@@ -0,0 +1,89 @@
...
+      followsMe: false,
+    },
+    {
+      profile: preFabProfiles[2],
+      icon: preFabProfiles[2].icon,
+      name: preFabProfiles[2].name,
+      following: false,
+      followsMe: true,
+    },
+  ];
```
Ajaxan, line 37: I would use the stored graph data in Redux instead of creating it here. Then, when we switch to the SDK the datatypes will match up and it will still work as opposed to requiring changes to the Profile to hook it up.|

There are a few changes that will be needed for the redux store to make sure we populate correctly with mock data though. Like right now we load in a list of people you follow and are following but we aren't grabbing their profiles. It'd make sense to load up the profile of everyone in your graph in the background. 

Let me know if you want to pair working on that.

src/components/FollowersFollowing.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { useState } from "react";
+import { Button } from "antd";
+import FollowersFollowingUsers from "./FollowersFollowingUsers";
+import { preFabProfiles } from "../test/testProfiles";
+
+const FollowersFollowing = (): JSX.Element => {
```
Ajaxan, line 6: This name is a bit confusing. My first though is that this was a component that was monitoring the followers that are following. And made me wonder where the FolloweesFollowing component was.

I think a name like `ProfileFollows` would be clearer. It's the component that handles all follows for the profile.

src/components/test/FollowersFollowing.test.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import FollowersFollowing from "../FollowersFollowing";
+import { shallow } from "enzyme";
+
+describe("Feed", () => {
```
Ajaxan, line 5: Feed

src/components/test/FollowersFollowing.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+  it("renders without crashing", () => {
+    expect(() => {
+      shallow(<FollowersFollowing />);
+    }).not.toThrow();
+  });
+  it("to match snapshot", () => {
+    //write more specific tests when we are pulling from a real followers/following list
```
Ajaxan, line 12: After integrating with the redux graph (see above comment) those test can be written.

src/components/test/FollowersFollowingUsers.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }).not.toThrow();
+  });
+  it("to match snapshot", () => {
+    const component = mount(
+      <FollowersFollowingUsers
+        selectedListTitle="showFollowers"
+        tempUserList={mockTempUserList}
+      />
+    );
+    expect(component).toMatchSnapshot();
```
Ajaxan, line 48: I'd like to try avoiding using `MatchSnapshot()` where possible. The issue with it is that every time we change the component we have to re-examine the whole snapshot to make sure the test is still working as intended. And if you ever miss something looking back through the snapshot the test stops working as intended.

src/components/test/FollowersFollowingUsers.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  {
+    profile: preFabProfiles[2],
+    icon: preFabProfiles[2].icon,
+    name: preFabProfiles[2].name,
+    following: false,
+    followsMe: true,
+  },
+];
+
+describe("Feed", () => {
```
Ajaxan, line 30: Feed

src/components/test/ProfileBlock.test.tsx
```diff
@@ -3,13 +3,66 @@ import { mount } from "enzyme";
...
+      const component = mount(componentWithStore(ProfileBlock, store));
+      component.find(".Profile__editButton").first().simulate("click");
+      component
+        .find(".Profile__name")
+        .simulate("change", { target: { value: "Monday NewLastName" } });
+      expect(
+        component.find(".Profile__editButton").first().props().disabled
+      ).toEqual(false);
+    });
+    // it("saves on save button click", () => {
```
Ajaxan, line 48: remove?

Ajaxan (changes requested):
Overall looks good, biggest things I'd suggest are to: 
1. Integrate with Redux now as opposed to later so the SDK integration will be easier
2. Not rely on Snapshots for tests since they can easily stop doing their job if we ever have to regenerate them and don't examine them fully.

Down to pair on the Redux changes if you'd like.

src/components/FollowersFollowing.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { Profile } from "../utilities/types";
+import { ListStatus } from "../utilities/enums";
+
+const FollowersFollowing = (): JSX.Element => {
+  const graph = useAppSelector((state) => state.user.graph);
+  const [selectedListTitle, setSelectedListTitle] = useState<ListStatus>(
+    ListStatus.CLOSED
+  );
+
+  const [usersList, setUsersList] = useState<Profile[] | any>([]);
```
Ajaxan, line 15: I believe the state for this should always be only `Profile[]`. Adding `any` subverts adding a typing.

I'd consider renaming this to `selectedList` to correspond with `selectedListTitle`. It helps clarify it's purpose. When I first read this variable I thought it was a list for the user which was confusing.

src/components/test/UserAvatar.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+  });
+  it("renders without crashing", () => {
+    const component = shallow(
+      <UserAvatar profile={getPrefabProfile(0)} avatarSize="large" />
+    );
+    expect(component).toMatchSnapshot();
+  });
```
Ajaxan, line 15: Name doesn't match test

src/components/test/FollowersFollowingUsers.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+      shallow(
+        <FollowersFollowingUsers
+          listStatus={ListStatus.FOLLOWERS}
+          tempUserList={mockTempUserList}
+          notFollowing={[preFabProfiles[0]]}
+        />
+      );
+    }).not.toThrow();
+  });
+});
```
Ajaxan, line 26: Should probably add a test here to make sure that the follow/unfollow is correctly being assigned based on the notFollowing and userList interaction.

src/components/test/FollowersFollowing.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+  it("renders without crashing", () => {
+    const profile = getPrefabSocialAddress(0);
+    const graph = getPreFabSocialGraph().get(profile);
+    const userGraph = { user: { graph } };
+    const store = createMockStore(userGraph);
+    expect(() => {
+      mount(componentWithStore(FollowersFollowing, store));
+    }).not.toThrow();
+  });
+});
```
Ajaxan, line 17: A test or two here testing to make sure the proper data is being passed to FollowersFollowingUsers as well as the proper list title may be a good idea.

src/utilities/enums.ts
```diff
@@ -0,0 +1,5 @@
+export enum ListStatus {
+  CLOSED,
+  FOLLOWERS,
+  FOLLOWING,
+}
```
Ajaxan, line 5: Even though it's used in a few different places I would keep this in the FollowersFollowing component. Since it's specific to that component and only used by directly adjacent components. I feel we should use the utilities for types that are used all over the place.

src/theme/_typography.scss
```diff
@@ -4,14 +4,22 @@ $font-sans-serif: -apple-system, blinkmacsystemfont, "Segoe UI", "Roboto",
...
+@mixin medium-text {
+  font-size: 18px;
+}
+
+@mixin small-text {
+  font-size: 13px;
+}
+
+@mixin bold-text {
+  font-weight: 550;
```
Ajaxan, line 24: If I'm not mistaken these generalized types are where we should start creating the kss styleguide.

src/index.scss
```diff
@@ -1,32 +1,7 @@
...
-//
-// Markup:
-// <button class="btn btn-bold">Click Me</button>
-//
-// Styleguide Components.Buttons.bold
-.btn.btn-bold {
-  position: relative;
-  font-weight: bold;
-  text-transform: uppercase;
-}
```
Ajaxan, line 24: This can't be removed until we add in some kss styleguide documentation elsewhere. Otherwise kss will fail on generation.

Ajaxan (changes requested):
It's looking good! Few more redux additions and tests for components and it should be all set

src/components/FollowersFollowing.tsx
enddynayn, line 15: its been a while since using Redux, does this belong in the Redux store?

claireolmstead: i like this. thanks!

src/utilities/enums.ts
claireolmstead: @wilwade what do you think about this?

src/components/FollowersFollowing.tsx
claireolmstead: this is local to our component and its child so I think we can use state here.

src/theme/_typography.scss
claireolmstead: new story for that?

src/utilities/enums.ts
claireolmstead: @wilwade what do you think?

claireolmstead:
waiting on session storage to be merged in.

src/components/ConnectionsList.tsx
```diff
@@ -0,0 +1,138 @@
...
+  ) => {
+    const followingProfiles: Profile[] = await Promise.all(
+      (following || []).map(
+        async (socialAddress: string) =>
+          await getConnectionProfile(socialAddress)
+      )
+    );
+    const followersProfiles: Profile[] = await Promise.all(
+      (followers || []).map(
+        async (socialAddress: string) =>
```
enddynayn, line 48: I think you can remove the async

src/components/test/UserAvatar.test.tsx
Ajaxan: I think we can remove the resetAllMocks here since we have no mocks.

src/components/test/UserAvatar.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+  it("renders without crashing", () => {
+    expect(() => {
+      shallow(<UserAvatar profile={getPrefabProfile(0)} avatarSize="large" />);
+    }).not.toThrow();
+  });
+  it("matches snapshot", () => {
+    const component = shallow(
+      <UserAvatar profile={getPrefabProfile(0)} avatarSize="large" />
+    );
+    expect(component).toMatchSnapshot();
```
Ajaxan, line 19: Would like to see tests other than snapshot whenever possible since they break more easily.

src/services/session.ts
```diff
@@ -1,8 +1,9 @@
-import { Profile } from "../utilities/types";
```
Ajaxan, line 1: Still grabbing my old commits for session storage story. These commits are outdated and adding them back in would break things.

src/index.scss
```diff
@@ -17,16 +20,48 @@ body {
...
+//  font-size: 18px;
+//}
+//
+//@mixin small-text {
+//  font-size: 13px;
+//}
+//
+//@mixin bold-text {
+//  font-weight: 550;
+//}
```
Ajaxan, line 67: Do we need these comments?

src/utilities/types.d.ts
```diff
@@ -42,4 +42,4 @@ export interface Graph {
   followers: HexString[];
 }
 
-export type SocialGraph = Map<HexString,Graph>;
+export type SocialGraph = Array<HexString, Graph>;
```
Ajaxan, line 45: I've never seen Array used with value pairs. Why use it over Map?

src/utilities/enums.ts
Ajaxan: Still think that splitting up this into it's own entire file just to see it be used in two related components seems unnecessary. It'd be cleaner if the Parent component of the two made it and the Child imported it.

src/components/ConnectionsList.tsx
```diff
@@ -0,0 +1,144 @@
+import React, { useEffect, useState } from "react";
+import { Button } from "antd";
+import ConnectionsListProfiles from "./ConnectionsListProfiles";
+import { useAppDispatch, useAppSelector } from "../redux/hooks";
+import { getProfile } from "../services/sdk";
```
Ajaxan, line 5: One thing I'd like to start doing for the sake of readability is using `* as sdk` for the sdk and also similar for our other internal APIs/services. With the goal of making it easier to navigate. Instead of having to scroll up to the top just to see where a method comes from, you can immediately tell. 

Examples would be things like:
`import * as sdk from "../services/sdk";`
`import * as wallet from "../services/wallet";`
`import * as session from "../services/session";`

And now our code, instead of having a random `getProfile` method that you now need to scroll up and figure out where it's from you can easily tell because it will become `sdk.getProfile`. Which makes it much more quickly apparent where it's from. It's slightly more verbose, but it's easier to follow which is our main goal.

Ajaxan (changes requested):
Few changes to make, but hopefully we can get this merged in soon!

src/components/test/UserAvatar.test.tsx
claireolmstead: @Ajaxan maybe we just remove the snapshot altogether and just leave it as does not throw an error.  Not really sure what else to test.

src/components/ConnectionsList.tsx
claireolmstead: I like this idea a lot as far as readability. The only argument I have against it is that by importing with the `*` you are importing all exports from that file, which is unnecessary when you are only using, as in this case, one or two of the functions. Is it worth it? I'm not sure.

src/services/session.ts
claireolmstead: **note: session storage broke in Josh's PR.  Bug fix story in pivotal.

Ajaxan (approved):
👍

src/components/ConnectionsListProfiles.tsx
```diff
@@ -0,0 +1,64 @@
...
+enum ListStatus {
+  CLOSED,
+  FOLLOWERS,
+  FOLLOWING,
+}
+
+interface ConnectionsListProfilesProps {
+  listStatus: ListStatus;
+  connectionsList: Profile[];
+  notFollowing: Profile[];
```
enddynayn, line 15: maybe we can call it `notFollowingProfiles`

src/components/ConnectionsListProfiles.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+enum ListStatus {
+  CLOSED,
+  FOLLOWERS,
+  FOLLOWING,
+}
+
+interface ConnectionsListProfilesProps {
+  listStatus: ListStatus;
+  connectionsList: Profile[];
```
enddynayn, line 14: is this a list of profiles that I am connected with or is this all of the social graph?

src/components/ConnectionsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const cachedProfiles: Profile[] = useAppSelector(
+    (state) => state.profiles.profiles
+  );
+  const [selectedListTitle, setSelectedListTitle] = useState<ListStatus>(
+    ListStatus.CLOSED
+  );
+  const [selectedList, setSelectedList] = useState<Profile[]>([]);
+  const [followingList, setFollowingList] = useState<Profile[]>([]);
+  const [followersList, setFollowersList] = useState<Profile[]>([]);
+  const [notFollowing, setNotFollowing] = useState<Profile[]>([]);
```
enddynayn, line 32: for naming consistency perhaps notFollowingList

src/components/ConnectionsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const dispatch = useAppDispatch();
+
+  const getConnectionProfile = async (
```
enddynayn, line 35: getConnectionProfile for a name is a little misleading because it also upserts a profile.  I think decoupling the getting and upsert will make thing a little more clear.

src/components/ConnectionsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+  };
+
+  const getUserConnectionsList = async (
+    following: HexString[],
+    followers: HexString[]
+  ) => {
+    const followingProfiles: Profile[] = await Promise.all(
+      (following || []).map(
+        async (socialAddress: string) =>
+          await getConnectionProfile(socialAddress)
```
enddynayn, line 55: I think we can drop the await

src/components/ConnectionsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const getClassName = (name: ListStatus) => {
+    return selectedListTitle === name
+      ? "ConnectionsList__button ConnectionsList__button--active"
+      : "ConnectionsList__button";
+  };
+
+  return (
+    <div className="ConnectionsList__block">
+      <div className="ConnectionsList__buttonBlock">
+        <Button
```
enddynayn, line 118:
```
    <Button
          className={getClassName(ListStatus.FOLLOWERS)}
          onClick={() => handleClick(ListStatus.FOLLOWERS)}
        >
          <div className="ConnectionsList__buttonCount">
            {graph?.followers.length}
          </div>
          Followers
        </Button>
        ```
        maybe you can make this a small component

File: src/components/ConnectionsList.tsx (hunk 41 above)
Comment (claireolmstead) [reply src/components/ConnectionsList.tsx:32]: i like this

File: src/components/ConnectionsListProfiles.tsx (hunk 39 above)
Comment (claireolmstead) [reply src/components/ConnectionsListProfiles.tsx:15]: went for notFollowingList

File: src/components/ConnectionsListProfiles.tsx (hunk 40 above)
Comment (claireolmstead) [reply src/components/ConnectionsListProfiles.tsx:14]: just the list that you are either following or are following you

File: src/components/ConnectionsList.tsx (hunk 43 above)
Comment (claireolmstead) [reply src/components/ConnectionsList.tsx:55]: you dont! sweet!

File: src/components/ConnectionsList.tsx (hunk 44 above)
Comment (claireolmstead) [reply src/components/ConnectionsList.tsx:118]: I kind of like having it right here so you.  more readable instead of having to dig and figure out where everything is.

File: src/components/ConnectionsList.tsx (hunk 42 above)
Comment (claireolmstead) [reply src/components/ConnectionsList.tsx:35]: but the upsert is only one line of code and i think it is more readable to have there.  the order of operations is very clear compared to pulling it out into another function.

Comment (enddynayn) [reply src/components/ConnectionsList.tsx:35]: maybe rename to upsertConnectionProfile

File: src/components/ConnectionsList.tsx (hunk 44 above)
Comment (enddynayn) [reply src/components/ConnectionsList.tsx:116]: you can keep it in the same file

File: src/components/ConnectionsList.tsx [hunk 45]
```diff
@@ -0,0 +1,148 @@
...
+    }
+    return userProfile;
+  };
+
+  const getUserConnectionsList = async (
+    following: HexString[],
+    followers: HexString[]
+  ) => {
+    const followingProfiles: Profile[] = await Promise.all(
+      (following || []).map(async (socialAddress: string) =>
```
enddynayn, line 53: I think we can remove the async here

src/components/ConnectionsList.tsx
```diff
@@ -0,0 +1,148 @@
+import React, { useEffect, useState } from "react";
```
enddynayn, line 1: I think we are toggling here? Maybe another name to consider is ToggleConnectionList

wilwade (approved):
👍

PR #14, opened 2021-04-12
Feature/setup redux store 177598915

Ajaxan:
Purpose
---------------
As a developer we want to use redux to store our state globally to simplify the applications state management and prop passing.

Solution
---------------
Implement a redux store and fix the existing code to use that store instead of default react state.
with @claireolmstead 

Change summary
---------------
* Added a redux store using the redux & react-redux & redux-toolkit modules
* Created global state with user, graph, profile, and feed store slices.
* Removed react state passing and simplified components.
* Fixed tests to use redux by implementing the redux-mock-store module.

Steps to Verify
----------------
1. Log in
2. Current state data appears in profile block
3. If Redux extension installed check state in Elements inspector


src/redux/slices/graphSlice.ts
```diff
@@ -0,0 +1,50 @@
...
+  userAddress: HexString;
+  targetAddress: HexString;
+}
+
+export const graphSlice = createSlice({
+  name: "graphs",
+  initialState,
+  reducers: {
+    setGraph: (state, action: PayloadAction<Graph>) => {
+      const newGraph: Graph = action.payload;
```
wilwade, line 22: Note that types coming out of the payload can be inferred from PayloadAction.

src/App.tsx
```diff
@@ -1,88 +1,19 @@
...
 import { BrowserRouter as Router } from "react-router-dom";
 import "./App.scss";
 import "antd/dist/antd.less";
-import {
-  clearSession,
-  getSession,
-  hasSession,
-  setSession,
-  updateSession,
-} from "./services/session";
```
wilwade, line 11: This service can also be deleted as it is not used anymore.

src/redux/slices/userSlice.ts
```diff
@@ -0,0 +1,48 @@
...
+  profile: Profile | null;
+  graph: Graph | null;
+}
+
+const initialState: userState = {
+  profile: null,
+  graph: null,
+};
+
+export const userSlice = createSlice({
```
wilwade, line 14: I don't see anything in here that is storing the session into local/session storage. Is that happening automagically?

wilwade (commented):
Few small comments.

src/redux/slices/userSlice.ts
Ajaxan: I was going to do that in a separate story.

src/App.tsx
Ajaxan: do you mean the session.js file? If so I was planning to add the userSlice data to session storage in a different story. I was planning to use the existing session.js as a reference as a write it.

src/utilities/types.d.ts
```diff
@@ -1,15 +1,14 @@
...
+import { PersonActivityPub, ActivityPub } from "./activityPub";
 
 export type HexString = string;
 export type EncryptedString = string;
 export type URLString = string;
 // ### Feed Data Types ###
-export type ContentType = "Person" | "Note" | "EncryptedMessage";
 
 // ## Profile ##
 export interface Profile extends PersonActivityPub {
```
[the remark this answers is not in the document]
Ajaxan: Does it make sense to have all of that fall under the "Profile" umbrella? At least then we can avoid specific implementation of stuff inside profiles but know they'll contain the handle/address/id and whatever else is needed.

src/redux/slices/userSlice.ts
```diff
@@ -0,0 +1,48 @@
...
+  name: "user",
+  initialState,
+  reducers: {
+    userLogin: (state, action: PayloadAction<userState>) => {
+      const profile: Profile = action.payload.profile as Profile;
+      const graph: Graph = action.payload.graph as Graph;
```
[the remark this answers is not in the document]
Ajaxan: The userState type is `type | null` since the user can be logged out and we don't know their profile or graph. If there's a better way to do this I can change it.

src/redux/slices/userSlice.ts
Ajaxan: https://www.pivotaltracker.com/story/show/177602937

src/redux/slices/userSlice.ts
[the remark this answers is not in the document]
wilwade: Ah yes there is. So in this case if null were to come in, it would just get overlooked until it threw an NPE.

In this case you never expect that to happen, so the first option might be to check for null and throw if they are. That will fix the types afterward as TS is smart enough to infer that the types are no longer nullable:

```suggestion
      const profile = action.payload.profile;
      const graph = action.payload.graph;
      if (profile === null || graph === null) throw new Error("Login Failure");
```

(Note that error is just an example one).

Another way would be to create a version of `userState` that didn't allow for nulls in the first place. That way when you reduce `userLogin`, the parameters cannot be null. There are two utility types around this, but neither of them quite work. One is `NonNullable`, but it is for the type only, not the nested interface. Another utility type for it is https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype but it only works for optionals instead of nullable. So an option would be to swap from null to optionals:

```suggestion

Interface UserState {
  profile?: Profile;
  graph?: Graph;
}

const initialState: UserState = {};

    userLogin: (state, action: PayloadAction<Required<UserState>>) => {
      const profile = action.payload.profile;
      const graph = action.payload.graph;
```

Also side node: types tend to be PascalCase.

src/utilities/types.d.ts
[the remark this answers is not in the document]
wilwade: I'm not sure yet? ActivityPub Profiles might be less important with handles.

src/redux/slices/userSlice.ts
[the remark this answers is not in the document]
Ajaxan: I went with optionals since it looks clean and readable plus throwing an error in a reducer felt bad.

src/utilities/types.d.ts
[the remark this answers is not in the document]
Ajaxan: I wasn't meaning to referring to the Activity pub aspect, just that their is a specific amount of data associated with our user that could logically fall under the idea of their "Profile". Even if it's just their Wallet Address, DSNP ID, and DSNP handle.

src/redux/slices/graphSlice.ts
```diff
@@ -0,0 +1,50 @@
...
+      state.graphs.get(socialAddress)?.following.push(followAddress);
+      state.graphs.get(followAddress)?.followers.push(socialAddress);
+      return state;
+    },
+    unfollow: (state, action: PayloadAction<AddressHolder>) => {
+      const socialAddress = action.payload.userAddress;
+      const unfollowAddress = action.payload.targetAddress;
+      const following = state.graphs.get(socialAddress)?.following;
+      const followers = state.graphs.get(unfollowAddress)?.followers;
+      following?.splice(following.indexOf(unfollowAddress));
```
enddynayn, line 43: are we modifying state here?

Ajaxan: I believe so. `splice` modifies the array it is called on, so this modifies the following and followers arrays.

package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@reduxjs/toolkit": "^1.5.1",
     "@toruslabs/torus-embed": "1.9.2",
-    "@types/node-fetch": "^2.5.8",
+    "@types/node-fetch": "^2.5.10",
```
enddynayn, line 8: why are we using node fetch?

Ajaxan: We were using it for easier fetch with the server in friendly. We were planning to use it to do the same with the example client server. However it currently isn't being used. I'll remove it for now, and just add it back when we need it.

enddynayn (approved):

enddynayn:
fabulous 👍

src/test/testGraphs.ts
```diff
@@ -1,31 +1,44 @@
-import { Graph, HexString } from "../utilities/types";
+import { Graph, HexString, SocialGraph } from "../utilities/types";
 import { generateSocialAddress, getPrefabSocialAddress } from "./testAddresses";
-import { randInt } from "./testhelpers";
+
+export const generateRandomGraph = (
+  socialAddress: HexString,
+  size: number = 4
+): Graph => {
+  const following: HexString[] = [];
```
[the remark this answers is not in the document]
Ajaxan: Ooh, that's much better

PR #16, opened 2021-04-14
converted all js files in src to ts with types

Ajaxan:
Purpose
---------------
We want to have the entirety of our src written in typescript but some js files slipped through without being converted.

Solution
---------------
Convert files in src from js to ts.

Change summary
---------------
* Converted and added types for torus wallet files
* Converted session storage file and added types.

Steps to Verify
----------------
1. Check all files in src
2. No js files should exist
3. Example client should still run


src/services/wallets/torus.ts
```diff
@@ -0,0 +1,64 @@
...
+  return await web3Torus.torus?.getPublicAddress({
+    verifier,
+    verifierId,
+    isExtended: true,
+  });
+};
+
+export const getWalletAddress = async (): Promise<string> => {
+  const { verifier, verifierId } = await getUserInfo();
+  const walletObject = await getPublicAddress(
```
enddynayn, line 51: the return value of this is an object or a string?
why not call this publicAddress?

src/services/wallets/torus.ts
```diff
@@ -0,0 +1,64 @@
...
+    verifier as VerifierTypes,
+    verifierId
+  );
+  if (!walletObject) throw new Error("No Wallet Address found for User");
+  else if (typeof walletObject === "string") return walletObject;
```
enddynayn, line 56: You can drop else from this and do

```
if (typeof walletObject === "string") return walletObject;

return walletObject.address;

```

src/services/wallets/tweb3.ts
```diff
@@ -0,0 +1,82 @@
...
+  //tncLink?: LocaleLinks<string>
+  //privacyPolicy?: LocaleLinks<string>
+  //contactLink?: LocaleLinks<string>
+  //customTranslations?: LocaleLinks<any>
+};
+
+// interface TorusParams in trous-embeded types
+const initSettings = (buildEnv: BuildEnvironment) => {
+  return {
+    buildEnv: buildEnv || "production",
```
enddynayn, line 45: maybe we can have a config file?

src/services/wallets/tweb3.ts
```diff
@@ -0,0 +1,82 @@
...
+    const web3Inst = new Web3(provider);
+    web3Torus.web3 = web3Inst;
+    this.web3.setProvider(provider);
+  },
+  initialize: async function (buildEnv: BuildEnvironment): Promise<void> {
+    const torus = new Torus(torusSettings);
+    await torus.init(initSettings(buildEnv));
+    this.initialized = true;
+    try {
+      await torus.login({ verifier: undefined });
```
enddynayn, line 70: why do we pass in undefined?

src/services/wallets/tweb3.ts
```diff
@@ -0,0 +1,82 @@
...
+      web3Torus.setweb3(torus.provider);
+      web3Torus.torus = torus;
+      sessionStorage.setItem("pageUsingTorus", buildEnv || "");
```
enddynayn, line 73: maybe we can move this out of here so we can isolate the responsibility of this method.

src/services/wallets/torus.ts
Ajaxan: It returns the wallet address of the user.

Since we have both a social identity address generated by the DSNP and a blockchain generated user/wallet address which are both public we need to be more specific. So we have opted to use `socialAddress` and `walletAddress` respectively to refer to these two different strings of characters.

src/services/wallets/tweb3.ts
Ajaxan: yeah that should definitely be done at some point. I'm going to add a story about adding a config in.

src/services/wallets/tweb3.ts
Ajaxan: Ah yeah it's a bit annoying. The login function requires you to pass an object with a verifier property but then doesn't require that the verifier have a real value. 

Anyway we don't want/need a verifier for the login so I just pass it undefined.

src/services/wallets/tweb3.ts
Ajaxan: That's the plan moving forward in the sessionStorage and wallet API stories. The current wallet in use will be stored in the user's data which will be kept in sessionStorage.

For this story though, I didn't want to mess with it since it's outside the scope.

src/services/wallets/torus.ts
shannonwells: Agree this is more "functional" style

shannonwells (commented):
Nothing to add

src/services/wallets/tweb3.ts
enddynayn: Sometimes it is necessary to merge things in fast, is this is one of those cases?

If we merge this in we are creating more code debt. This PR is a result of code debt and I think it's probably best we don't create more. 

It might be better if we make the update now since it would probably take more effort to write up a story and plan for it.

Let me know if you wanna pair.

😄

Ajaxan: It's not about getting this one in fast that code was there already. Actually this line has been around for awhile, It just looks new because the file was converted to typescript.

My thinking is that while this does need to be fixed so does session storage as a whole. That story has already been planned and created and is bigger than this small piece. Since that's already going to need to be done regardless, this will be fixed at that point. 

Now the code should work if I just delete the line but then I might forget to integrate tracking of Torus into session storage later. Relying on my ability to remember seems worse than leaving it this way till the official story to change it comes up.

enddynayn: Oh I see. I think it is better to delete it then if its not needed. You can make a note on the ticket as a reminder.

src/services/wallets/torus.ts
enddynayn: I think writing address is ok here. WalletObject is more confusing.

src/services/wallets/tweb3.ts
Ajaxan: Yeah I can do that. I'll delete it here and add a comment in the story so I don't forget

enddynayn (approved):

PR #17, opened 2021-04-16
Added .env file with basic vars

Ajaxan:
Purpose
---------------
As a developer I want to have easily changed environmental variables to make testing and running the example client easier.

Solution
---------------
Add .env file with some basic variables for the chain and startup port

Change summary
---------------
* startup port is now configurable
* chain connection vars for torus are now configurable

Steps to Verify
----------------
1. Login with Torus
2. Should connect and obtain WalletAddress


src/services/wallets/tweb3.ts
```diff
@@ -42,11 +42,12 @@ const popupStyling = {
...
-    buildEnv: buildEnv || "production",
+    buildEnv:
+      (process.env.BUILD_ENV as BuildEnvironment) || buildEnv || "production",
     network: {
-      host: "http://localhost:7545",
-      chainId: 1337,
-      networkName: "Localchain",
+      host: process.env.CHAIN_HOST || "http://localhost:7545",
+      chainId: Number(process.env.CHAIN_ID) || 1337,
+      networkName: process.env.CHAIN_NAME || "Localchain",
```
enddynayn, line 50: maybe keep these in one place an import a config object?

Ajaxan: I thought about making a JSON config file and just importing it as an object. I ended up going with the .env file since it seemed to be more widely used for these types of global constants. I also didn't want to create them to be specific to Torus. It's likely that later on things like CHAIN_ID and CHAIN_HOST may need to be used elsewhere.

.env
```diff
@@ -0,0 +1,7 @@
+PORT=3000
```
wilwade, line 1: Usually these are not committed and it is the same as the defaults you have in the code. Thoughts?

.env
```diff
@@ -0,0 +1,7 @@
...
+
+CHAIN_ID=1337
+CHAIN_HOST=http://localhost:7545
+CHAIN_NAME=Localchain
+
+BUILD_ENV=testing
```
wilwade, line 7: This sounds a lot like `test` which I don't think is what you mean? I would have expected `local` or `development`

wilwade: Ah now I see that this is torus specific I think? Everything that is just for torus should be prefixed.

.env
```diff
@@ -0,0 +1,7 @@
+PORT=3000
+
+CHAIN_ID=1337
```
wilwade, line 3: I do not believe these will actually work without a prefix: https://create-react-app.dev/docs/adding-custom-environment-variables/

wilwade (changes requested):

.env
Ajaxan: I was looking up the docs for the `.env` file and they did recommend committing it. And if you want to set different settings create a `env.local` which overrides the `.env` file. 

https://create-react-app.dev/docs/adding-custom-environment-variables/

As for it already being the same as the default. I added it to the file so it's more clear that it can be changed. If you think it makes more sense to remove it though I'll do that.

.env
Ajaxan: Yeah the build environment in this case was Torus specific, I'll make that more clear.

.env
Ajaxan: You're right, there were a few other issues I fixed as well.

src/services/wallets/torus/tweb3.ts
```diff
@@ -67,6 +70,9 @@ const web3Torus = {
     await torus.init(initSettings(buildEnv));
     this.initialized = true;
     try {
+      console.log("CHAIN_ID: ", process.env.REACT_APP_CHAIN_ID);
```
enddynayn, line 73: left on accident

enddynayn (approved):

package.json
```diff
@@ -25,6 +27,7 @@
     "react-dom": "17.0.1",
     "react-redux": "^7.2.3",
     "react-router-dom": "^5.2.0",
+    "react-scripts": "^4.0.3",
```
wilwade, line 30: @Ajaxan I don't think we need this do we? Since it is ejected.

src/App.tsx
```diff
@@ -5,6 +5,7 @@ import "antd/dist/antd.less";
 import Header from "./components/Header";
 import Feed from "./components/Feed";
 import ProfileBlock from "./components/ProfileBlock";
+//require("dotenv").config();
```
wilwade, line 8: Comment.

wilwade (changes requested):
Two quick final questions. Otherwise looking good!

wilwade (approved):
🚄

PR #18, opened 2021-04-20
Feature/create wallet api 177602979

Ajaxan:
Purpose
---------------
As a developer I want a singular wallet API to integrate into the example client as opposed to having to worry about separate implementations for multiple wallets.

Solution
---------------
Create a singular wallet API that implements common methods we need from a wallet and handles the specific communication with those wallets and their APIs.

Change summary
---------------
* Created Wallet api interface that all wallet's implements
* Added current wallet to userSlice of redux
* Implemented Metamask wallet to demonstrate wallet api
* Added second login option for Metamask
* Fixed tests to account for new interface

Steps to Verify
----------------
1. Login with Torus
2. logout
3. login with Metamask
4. logout
5. All steps should work. And verify with wallet address


src/redux/slices/userSlice.ts
```diff
@@ -15,8 +17,10 @@ export const userSlice = createSlice({
     userLogin: (state, action: PayloadAction<UserState>) => {
       const profile = action.payload.profile;
       const graph = action.payload.graph;
+      const wallet = action.payload.wallet;
       state.profile = profile;
       state.graph = graph;
+      state.wallet = wallet;
```
wilwade, line 23: Not really related, but why not:
```suggestion
      state.profile = action.payload.profile;
      state.graph = action.payload.graph;
      state.wallet = action.payload.wallet;
```

src/services/wallets/wallet.ts
```diff
@@ -0,0 +1,72 @@
...
+  switch (walletType) {
+    case WalletType.TORUS:
+      return torusWallet;
+    case WalletType.METAMASK:
+      return metamaskWallet;
+    // Add new WalletTypes Here
+  }
+};
+
+const torusWallet: Wallet = {
```
wilwade, line 31: I might shift this out to a different file perhaps `./metamask/index.ts`. That file would then export default just the wallet. Then everything else metamask is hidden inside the ./metamask module folder.

wilwade (commented):
Two quick notes on org from a quick look. Looking really good!

src/services/wallets/metamask/ethereum.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  request(request: RequestArguments): Promise<any>;
+}
+// eip-1193
+const ethereum = (window as any).ethereum as
+  | EthereumProvider
+  | null
+  | undefined;
+
+//ethereum.autoRefreshOnNetworkChange = false;
```
claireolmstead, line 43: remove or keep for future use?

src/services/wallets/metamask/metamask.ts
```diff
@@ -0,0 +1,25 @@
+import Web3 from "web3";
+import { HexString } from "../../../utilities/types";
+import ethereum from "./ethereum";
+
+export const mmweb3 = new Web3(ethereum as any);
+
+export const isInstalled = (): boolean => {
+  return Boolean(ethereum && ethereum.isMetaMask);
```
claireolmstead, line 8:
```suggestion
  return Boolean(ethereum?.isMetaMask);
```

src/services/wallets/torus/tweb3.ts
```diff
@@ -28,8 +28,10 @@ const popupStyling = {
...
   },
-  logoDark: "https://www.projectliberty.io/assets/images/brand-icon.png",
-  logoLight: "https://www.projectliberty.io/assets/images/brand-icon.png",
+  logoDark:
+    "https://www.pngkit.com/png/full/119-1193080_technology-electronics-icons-png.png",
+  logoLight:
+    "https://www.pngkit.com/png/full/119-1193080_technology-electronics-icons-png.png",
   //topupHide?: boolean
   //featuredBillboardHide?: boolean
   //disclaimerHide?: boolean
```
claireolmstead, line 37: can we delete this?

claireolmstead (commented):
looks pretty clean and useable to me!

src/components/Login.tsx
```diff
@@ -1,42 +1,43 @@
...
       const graph = await sdk.getGraph(socialAddress);
-      dispatch(userLogin({ profile, graph }));
-      history.push("/");
+      dispatch(userLogin({ profile, graph, wallet: walletType }));
       startLoading(false);
     } catch (error) {
       setAlertError(error.toString());
     }
     startLoading(false);
+    setPopoverVisible(false);
```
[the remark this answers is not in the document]
Ajaxan: This helps make sure to close the popup after the login is finished or errors out. There are instances, like a failed login, where the popup will stay open without this line.

src/services/wallets/metamask/ethereum.ts
Ajaxan: re-enabled

src/services/wallets/torus/torus.ts
```diff
@@ -48,17 +56,32 @@ export const getPublicAddress = async (
...
   if (!walletAddressHolder) throw new Error("No Wallet Address found for User");
   if (typeof walletAddressHolder === "string") return walletAddressHolder;
   return walletAddressHolder.address;
 };
 
+const getTypedVerifier = (untypedVerifier: string): VerifierTypes => {
+  console.log(
+    "🚀 | file: torus.ts | line 67 | untypedVerifier",
+    untypedVerifier
+  );
```
wilwade, line 70: Log?

src/services/wallets/torus/tweb3.ts
```diff
@@ -28,8 +28,10 @@ const popupStyling = {
       torusGray2: "#FBF7F3",
     },
   },
-  logoDark: "https://www.projectliberty.io/assets/images/brand-icon.png",
-  logoLight: "https://www.projectliberty.io/assets/images/brand-icon.png",
+  logoDark:
+    "https://www.pngkit.com/png/full/119-1193080_technology-electronics-icons-png.png",
+  logoLight:
+    "https://www.pngkit.com/png/full/119-1193080_technology-electronics-icons-png.png",
```
wilwade, line 34: These urls feel weird. Is that really what we want?

src/services/wallets/wallet.ts
```diff
@@ -0,0 +1,71 @@
...
+  getBalance: async (walletAddress: HexString) => {
+    const stringBal = await metamask.getBalance(walletAddress);
+    return Number(stringBal);
+  },
+  getWeb3: () => {
+    return metamask.getWeb3();
+  },
+}; */
+
+// Add new Wallet interface here
```
wilwade, line 71: I think we can remove this, although a comment about how to add new wallets might be good.

src/services/wallets/wallet.ts
```diff
@@ -0,0 +1,71 @@
...
+  TORUS = "TORUS",
+  METAMASK = "METAMASK",
+  // Add new WalletTypes Here
+}
+
+export interface Wallet {
+  login: () => Promise<HexString>;
+  logout: () => void;
+  getAddress: () => Promise<HexString>;
+  getBalance: (walletAddress: HexString) => Promise<number>;
```
wilwade, line 16: Do we use getBalance?

src/services/wallets/metamask/metamask.ts
```diff
@@ -0,0 +1,25 @@
...
+export const getWalletAddress = async (): Promise<HexString> => {
+  const response = await ethereum?.request({ method: "eth_requestAccounts" });
+  if (!response) throw new Error("Metamask not installed");
+  return response.result ? response.result[0] : response[0];
+};
+
+export const getBalance = (account: HexString): Promise<string> => {
+  return mmweb3.eth.getBalance(account).then((balance) => {
+    return balance;
+  });
```
wilwade, line 20: `.then(balance => { return balance })` doesn't look like it is doing anything.

src/services/wallets/metamask/metamask.ts
```diff
@@ -0,0 +1,25 @@
+import Web3 from "web3";
+import { HexString } from "../../../utilities/types";
+import ethereum from "./ethereum";
+
+export const mmweb3 = new Web3(ethereum as any);
```
wilwade, line 5: This any is odd. What type does Web3(param) expect?

src/components/Login.tsx
```diff
@@ -52,19 +53,38 @@ const Login = (): JSX.Element => {
...
+            Log In
+            {loading && <Spin className="Login__spinner" size="small" />}
+          </Button>
+        </Popover>
       ) : (
         <Button
           className="Login__logOutButton"
           aria-label="Logout"
-          onClick={torusLogout}
+          onClick={() => logout()}
```
wilwade, line 87:
```suggestion
          onClick={logout}
```

src/services/wallets/torus/torus.ts
```diff
@@ -1,15 +1,15 @@
...
-};
-
-export const isInitialized = (): boolean => {
-  return web3Torus.initialized;
-};
+// Torus doesn't give us access to their internal verifier types
+// They also don't enforce their own verifier system consistently
+// This means we need to create our own and update it.
+// This verifiers array is our list of approved verifiers based
+// off their internal list.
```
wilwade, line 9: Could you add a link to where to find the internal list?

wilwade (changes requested):
Few small cleanups, but looking nice! I love that nice clean top level interface.

src/services/wallets/torus/tweb3.ts
Ajaxan: Don't think it will be long term but we needed something. If we have a link to an unfinished or liberty icon I can use I'll put that in.

src/services/wallets/wallet.ts
Ajaxan: No I left it in thinking we might end up using it. I can remove it though

src/services/wallets/torus/tweb3.ts
wilwade: @claireolmstead do you have something that could be used here?

src/services/wallets/metamask/metamask.ts
```diff
@@ -0,0 +1,25 @@
+import Web3 from "web3";
+import { HexString } from "../../../utilities/types";
+import ethereum, { EthereumProvider } from "./ethereum";
+
+export const mmweb3 = new Web3(ethereum as EthereumProvider);
```
wilwade, line 5: Hmm... the as here is covering up the case where ethereum is undefined or null. We should do that check instead of hiding it. (This might require throwing an error in the case of `getWeb3`)

wilwade (changes requested):
One final question.

src/services/wallets/metamask/metamask.ts
Ajaxan: Yeah this will error out currently if someone attempts to log in with Metamask but Metamask isn't installed. I can have it throw an error that Metamask isn't installed. It'll be caught be the login and display as an alert by the login button.

src/services/wallets/torus/tweb3.ts
Ajaxan: I replaced it with the public github logo. It's a bit small but it's better and ours.

src/services/wallets/metamask/metamask.ts
Ajaxan: Hmm I'm no 100% happy with my solution but it can no longer go through without being checked and throwing an error if it's null or undefined.

src/services/wallets/metamask/metamask.ts
```diff
@@ -0,0 +1,20 @@
+import Web3 from "web3";
+import { HexString } from "../../../utilities/types";
+import ethereum, { EthereumProvider } from "./ethereum";
+
+const mmweb3 = new Web3(ethereum as EthereumProvider);
```
wilwade, line 5: @Ajaxan have you tried this in a browser without metamask? I think it might blowup.
```suggestion
const mmweb3 = ethereum ? new Web3(ethereum) : null;
```

src/services/wallets/metamask/metamask.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const getWeb3 = (): Web3 => {
+  if (!ethereum) throw new Error("Metamask not installed");
```
wilwade, line 8: Should this check `isInstalled()` instead? `ethereum` can be more than just metamask

wilwade (changes requested):
One last double check.

src/services/wallets/metamask/metamask.ts
Ajaxan: No, it throws an error that is caught by the login sequence and display "Metamask not installed" in an error alert next to the login button.

src/services/wallets/metamask/metamask.ts
Ajaxan: I'll go ahead and use isInstalled for now, but are there other wallets that may use the EthereumProvider?

wilwade (approved):
🚂

PR #20, opened 2021-04-21
use test generators package 177633808

acruikshank:
Work towards https://www.pivotaltracker.com/story/show/177633808

Purpose
---------------
We are starting to build common test functionality across multiple repos, so we've created an @unfinishedlabs/test-generators package to distribute this common functionality.

Solution
---------------
This PR replaces the generict test-generator functionality in sdk-ts with function from the test-generators package.

Change summary
---------------
* Import @unfinishedlabs/test-generators into the project.
* Remove common functions from testhelpers.ts
* Replace usages of test functions in testAddresses, testFeeds, testGraphs, and testProfiles

Steps to Verify
----------------
1. `npm run build` finishes without error
2. `npm run test` produces no test errors

Notes
----------------
- There are many test generator functions remaining in this project. We might want to move some of these into the test-generators package, but there is currently no common need for them.


Ajaxan (approved):
Good idea to make this package. Looks good to me!

PR #21, opened 2021-04-23
Feature/new post modal/177863520

claireolmstead:
Purpose
---------------
https://www.pivotaltracker.com/story/show/177863520
As a user, I want to be able to create a new post.  In order to do this I need modal to input my content.

Solution
---------------
Solution:
- create new post modal
- style new post modal

Change summary
---------------
Acceptance Criteria:
- use design from friendly and antd to style modal
- rather than saving data, just set state

- user can enter text, and/or a URL to a photo or video.
- user can attach multiple URLs to create a carousel
- as long as post content !== null, create post
- if content === null, disable post button

Steps to Verify
----------------
1. make sure that it looks good
2. saves correct information

----------------
Screen Shot 2021-06-24 at 3 43 44 PM


src/components/PostList.tsx
```diff
@@ -1,9 +1,37 @@
 import React from "react";
+import { useAppSelector } from "../redux/hooks";
+import { FeedItem } from "../utilities/types";
+import { NoteActivityPub } from "../utilities/activityPubTypes";
 
 const PostList: React.FC = () => {
+  const feed: FeedItem[] = useAppSelector((state) => state.feed.feed);
+  feed.filter((post) => post?.content?.type === "Note");
```
enddynayn, line 8: It looks like this is not being set to a variable so its not being used.

I think you meant to do something like
```
  const feed: FeedItem[] = 
     useAppSelector((state) => state.feed.feed).filter((post) => post?.content?.type === "Note")
 ```

File: src/components/PostList.tsx [hunk 2]
```diff
@@ -1,9 +1,37 @@
...
+
+  const postItems = feed.map((post: FeedItem, index: number) => {
```
enddynayn, line 10: I think we can refactor the individual postItem into a their own component.
```
const PostItem = ({fromAddress, noteContent, attachement, ....}) => (
  <div>
  </div>
);
```

src/components/Feed.tsx
```diff
@@ -1,13 +1,31 @@
-import React from "react";
+import React, { useState } from "react";
 import NewPost from "./NewPost";
 import PostList from "./PostList";
+import { Button } from "antd";
+import { useAppSelector } from "../redux/hooks";
 
 const Feed: React.FC = () => {
+  const profile = useAppSelector((state) => state.user.profile);
+  const [openCreatePost, setCreatePost] = useState<boolean>(false);
```
enddynayn, line 9: `setCreatePost` is a little misleading since all this is doing is opening the modal. Perhaps, isModalOpen?

src/components/PostList.tsx
claireolmstead: From checking out the snapshot, it seems like the opposite.  Maybe we can pair on this.

src/components/scss/Profile.css
```diff
@@ -0,0 +1,17 @@
+/* Error: Undefined variable.
+ *   ,
+ * 2 |   background-color: $gray;
+ *   |                     ^^^^^
+ *   '
+ *   Profile.scss 2:21  root stylesheet */
+
+body::before {
```
wilwade, line 8: We need to add `src/**/*.css` to our `.gitignore`.

src/components/PostList.tsx
```diff
@@ -1,9 +1,37 @@
...
+      <div key={index}>
+        {noteContent && (
+          <>
+            <div>{post.fromAddress}</div>
+            <div>{noteContent.content}</div>
+            {noteContent.attachment && (
+              <img
+                width={400}
+                src={noteContent.attachment[0]?.url}
+                alt="undefined"
```
wilwade, line 22: Why does `alt="undefined"` here? I would have expected either `""` or perhaps `=noteContent.attachment[0]?.url ? "Image for post" : ""` or something else. https://www.a11yproject.com/posts/2013-01-14-alt-text/

wilwade (changes requested):
Nothing much over what Enddy had.

The big thing is that we need to add some gitignores for css and css map files and clear those out.

src/components/PostList.tsx
claireolmstead: this is happening in another story.  the current code in postlist is just a placeholder.

src/components/NewPost.tsx
```diff
@@ -1,10 +1,95 @@
...
+  const success = () => {
+    setSaving(false);
+    setErrorMsg(null);
+    onSuccess();
+  };
+
+  const createPost = async () => {
+    //needs to store in batch file here
+    if (!profile) return;
+    const newPostFeedItem: FeedItem = await createNote(
```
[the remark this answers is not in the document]
claireolmstead: make it into a new chore story?

enddynayn (approved):

wilwade (approved):
🎉

PR #22, opened 2021-04-27
Feature/store user data in session 177602937

Ajaxan:
Purpose
---------------
As a user I want my data to be persistent as long as I have the page open. I don't want to have to re-login every time I refresh.

Solution
---------------
By storing our user login information and state from the Redux state in session storage we can make our global state persistent.

Change summary
---------------
* Created session storage
* Added user profile and wallet to session storage
* Added reload option for wallet api
* fixed tests for login and App
* Added icon for current logged in wallet by login button

Steps to Verify
----------------
1. Redux state is stored in sessionStorage
2. Login to example client
3. Hard refresh page
4. Should stay logged in
5. Close page
6. Reopen. should be logged out.


src/App.tsx
```diff
@@ -1,12 +1,19 @@
...
 import Feed from "./components/Feed";
 import ProfileBlock from "./components/ProfileBlock";
+import { wallet } from "./services/wallets/wallet";
+import { useAppSelector } from "./redux/hooks";
 
 const App: React.FC = () => {
+  const walletType = useAppSelector((state) => state.user.walletType);
+  useEffect(() => {
+    if (walletType) wallet(walletType).reload();
+  });
```
enddynayn, line 15: should wallet be passed in as a dependency?

src/redux/slices/profileSlice.ts
```diff
@@ -15,11 +15,18 @@ export const profileSlice = createSlice({
   reducers: {
     setProfile: (state, action: PayloadAction<Profile>) => {
       const newProfile = action.payload;
-      state.profiles.set(newProfile.socialAddress, newProfile);
+      const oldProfile = state.profiles.find(
+        (profile) => profile.socialAddress === newProfile.socialAddress
+      );
+      if (oldProfile) state.profiles.splice(state.profiles.indexOf(oldProfile));
```
enddynayn, line 21: do we want to remove a single item or multiple? Here I think splice is removing everything from the start.

src/redux/slices/profileSlice.ts
```diff
@@ -15,11 +15,18 @@ export const profileSlice = createSlice({
...
+      state.profiles.push(newProfile);
       return state;
     },
     removeProfile: (state, action: PayloadAction<HexString>) => {
-      state.profiles.delete(action.payload);
+      const oldProfile = state.profiles.find(
+        (profile) => profile.socialAddress === action.payload
+      );
+      if (oldProfile) state.profiles.splice(state.profiles.indexOf(oldProfile));
```
enddynayn, line 29: same here

src/services/session.ts
```diff
@@ -1,52 +1,28 @@
...
 
 export const clearSession = (): void => {
   sessionStorage.removeItem("session");
 };
 
-export const getSession = (): SessionData | null => {
+export const loadSession = (): SessionData | null => {
   try {
     const sessionData = sessionStorage.getItem("session");
+    console.log("🚀 | file: session.ts | line 14 | sessionData", sessionData);
```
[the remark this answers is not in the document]
Ajaxan: Whoops, thanks

src/services/wallets/metamask/index.ts
```diff
@@ -2,13 +2,18 @@ import * as metamask from "./metamask";
...
+  icon:
+    "https://cdn.iconscout.com/icon/free/png-512/metamask-2728406-2261817.png",
   login: async () => {
     if (!metamask.isInstalled()) throw new Error("Metamask not installed");
     return metamask.getWalletAddress();
   },
   logout: () => {
     return;
   },
+  reload: () => {
```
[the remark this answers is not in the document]
Ajaxan: It actually wasn't left out. The current way I'm implementing the Wallet API means I'm creating an interface that will have far more actions it will need to handle than any one wallet will likely implement. 

In this case wallets like Torus which are embedded in the site itself need to be able to reload while Wallets in the browser won't. As a result the function for implementing the reload for Metamask is empty.

src/App.tsx
Ajaxan: I don't believe so? We want the reload to trigger whenever the site reloads not just when the walletType changes.

src/redux/slices/profileSlice.ts
Ajaxan: It says for splice:
`"Removes elements from an array and, if necessary, inserts new elements in their place".`
So if I was adding multiple profiles yeah it would remove multiple but the array `find` says:
`"Returns the value of the first element in the array where predicate is true, and undefined otherwise."`

So I believe we should only ever find a single profile with the same address and replace it. My understanding is that the DSNP identity address is unique, so this should work as intended. Let me know if there may be an issue with my line of thinking on it.

enddynayn (approved):

src/redux/slices/profileSlice.ts
Ajaxan: Misread the question, I did make a mistake here, need to pass to splice the number of elements to remove starting at the passed index. Fixed, thanks!

PR #23, opened 2021-04-30
github pages added

claireolmstead:
Purpose
---------------
Publish the site so that people can test it and view progress without having to run it locally.

Solution
---------------
deploy with github pages

Change summary
---------------
* add gh-pages package
* add homepage and build instructions
* config pages to look at the gh-pages branch

Steps to Verify
----------------
1. Make sure the link works.

----------------
link to the published site:
https://libertydsnp.github.io/example-client/


src/App.css
```diff
@@ -0,0 +1,53 @@
+.Login__block .ant-btn {
```
Ajaxan, line 1: Remove .css files.

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "client",
+  "homepage": "http://LibertyDSNP.github.io/example-client",
```
Ajaxan, line 3: We still need a blockchain backend for this to work. Currently we've just been using the old demo contracts repo, but we can probably work on something new. Perhaps we can even include that in the repo? We should probably discuss it next week.

Ajaxan (changes requested):
We will need to setup a blockchain for the example client to fully get this up and working.

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "client",
+  "homepage": "https://libertydsnp.github.io/example-client/",
```
wilwade, line 3: This is just me, but I think the home page should still be the github repo.

PR #24, opened 2021-05-10
Feature/feed loading system 177893640

Ajaxan:
Purpose
---------------
As a developer I want a system that handles loading feed data from the sdk and supply both the new and old content as needed.

Solution
---------------
Create the required code to retrieve feed data and content and supply it to our redux store and components as needed.

Change summary
---------------
* Added new and old feed loading SDK API endpoints
* Added content loading server API endpoints
* Created mock endpoints for server and sdk as placeholders until functional
* Modified feed slice of redux store to handle feed data better
* Added unrefined feed loading implementation to components
* Fixed tests to include redux store

Steps to Verify
----------------
1. Start up example client
2. Feed should load
3. Content should appear for each post
4. Replies should be attached to their post
5. After a small delay "new content button" appears
6. Clicking button loads new feed posts
7. At any point check redux store for proper storage of data

----------------
image
image
image
image


src/redux/slices/feedSlice.ts
```diff
@@ -3,31 +3,140 @@ import { FeedItem } from "../../utilities/types";
 
 interface feedState {
   feed: FeedItem[];
+  newFeed: FeedItem[];
+  replies: FeedItem[];
+  currentBlock: number | null;
 }
```
Ajaxan, line 9: Would like some feedback on the way I'm storing/managing my feed data. I'm not 100% happy with the way it works. I feel it could be more efficient and perhaps a bit cleaner.

src/services/fakesdk.ts
```diff
@@ -25,3 +34,106 @@ export const getProfileFromSocialIdentity = async (
...
+  PrivateGraphKeyList = 5,
+  EncryptionKeyList = 6,
+  Reaction = 7,
+  PrivateGraphChange = 8,
+  Drop = 9,
+  EncryptedInbox = 10,
+  PrivateBroadcast = 11,
+  Reply = 12,
+  Batch = 13,
+}
```
Ajaxan, line 78: Same as above comment about ActionType and importing from the sdk.

src/components/Feed.tsx
```diff
@@ -1,13 +1,56 @@
...
 
 const Feed: React.FC = () => {
+  const [subscriptionID, setSubscriptionID] = React.useState<string>("");
+
+  const dispatch = useAppDispatch();
+  const newFeed = useAppSelector((state) => state.feed.newFeed);
+
+  useEffect(() => {
+    subscribe();
+    dispatch(addNewFeedtoMainFeed());
```
enddynayn, line 22: there is more than one feed? Like a personal feed or custom and public feed?

src/components/Feed.tsx
```diff
@@ -1,13 +1,56 @@
...
+    return unsubscribe;
+  }, []);
+
+  const subscribe = () => {
```
enddynayn, line 26: if you have time perhaps move this into a hook

src/components/PostList.tsx
```diff
@@ -1,10 +1,92 @@
...
+import * as server from "../services/server";
+import { addFeedItems } from "../redux/slices/feedSlice";
+
+const BLOCKS_PER_LOAD = 8;
 
 const PostList: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const feed = useAppSelector((state) => state.feed.feed);
+  let currentBlock: sdk.BlockNumber =
+    useAppSelector((state) => state.feed.currentBlock) ?? sdk.getNewestBlock();
```
enddynayn, line 16: what does a double ?? do?

src/components/PostList.tsx
```diff
@@ -1,10 +1,92 @@
...
+        next={loadMoreFeed}
+        hasMore={!currentBlock || currentBlock > 0}
+        loader={<h4>Loading...</h4>}
+        endMessage={
+          <p style={{ textAlign: "center" }}>
+            <b>End of Feed</b>
+          </p>
+        }
+      >
+        <List
```
enddynayn, line 64: can we please do a small refactor and create a FeedItem component in this file?

src/components/Feed.tsx
Ajaxan: I've made a distinction between feed data that comes in that is new and the main feed of already loaded data. Doing so makes it easy to allow the current loaded feed data to not be constantly pushed down by new incoming data. Instead that new feed data is loaded in when the page is refreshed or when the "load new content" button is pressed.

However beyond this single distinction there is no separating of feed data for personal or public. All of that would/will be done by filtering the existing feed.

src/components/Feed.tsx
Ajaxan: I actually made the entire hook to do this and then realized there wasn't any need for a hook at all and the subscriptionID state could be removed entirely. It's only needed to be passed to the unsubscribe which we don't need state for in this case.

PR #26, opened 2021-05-11
Feature/add none wallet type 178099866

Ajaxan:
Purpose
---------------
As a developer having a "none" `walletType` will allow us to deal with less nulls when checking for Wallets and their available data. Most wallet functions will not do anything and the "none" wallet type will effectively be the logged out state.

Solution
---------------
Added "none" wallet type that is the default state. No more dealing with union types for `walletType`. 

Change summary
---------------
* Changed redux slice state for walletType
* Changed walletType checks on login/logout

Steps to Verify
----------------
1) Start up example client, login button should appear
2) using login button works and it becomes logout
3) Logging out return button to login
4) You shouldn't be able to make posts or see followers and following while logged out. Should say "log in to see profile" or similar.

### DIFF
--- a/src/redux/slices/userSlice.ts
+++ b/src/redux/slices/userSlice.ts
@@ -5,14 +5,12 @@ import { Profile } from "../../utilities/types";
 
 interface UserState {
   profile?: Profile;
-  walletType?: wallet.WalletType;
+  walletType: wallet.WalletType;
 }
 
 const initialState: UserState = {
   profile: session.hasSession() ? session.loadSession()?.profile : undefined,
-  walletType: session.hasSession()
-    ? session.loadSession()?.walletType
-    : undefined,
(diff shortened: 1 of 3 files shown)


src/redux/slices/userSlice.ts
```diff
@@ -5,14 +5,12 @@ import { Profile } from "../../utilities/types";
...
-  walletType?: wallet.WalletType;
+  walletType: wallet.WalletType;
 }
 
 const initialState: UserState = {
   profile: session.hasSession() ? session.loadSession()?.profile : undefined,
-  walletType: session.hasSession()
-    ? session.loadSession()?.walletType
-    : undefined,
+  walletType: session.loadSession()?.walletType ?? wallet.WalletType.NONE,
```
enddynayn, line 13: aahhh....nullish coalescing, cool. super helpful.

enddynayn (approved):

PR #37, opened 2021-07-27
Retrieve Profiles from sdk (+ cleanup)

acruikshank:
Purpose
---------------
This PR kind of got away from me. There's a lot of good stuff in here. I can't really pull it apart into separate PRs at this point, so it's going to be kind of tough to review. Here are the key points:

Solution
---------------

1. The example client now pulls all log history from the SDK. This is blocked by a [bug](https://github.com/LibertyDSNP/sdk-ts/pull/107) in the SDK. Until this is resolved, the PR is using a locally imported version of the SDK.
2. I've added a `scripts/populate.mjs` script to create locally hosted content on chain. It currently creates a profile and note message from each account.
3. I've modified the example client to use SDK types where possible. It has been using it's own version of activity content types in test data, which were either speculative or old. This change allows us to use data from the SDK directly. This is responsible for most of the churn in the PR. Note: a related rename of `src/utilties/activityPub.ts` -> `src/utilties/activityContent.ts` is being tracked as deletion and creation rather than a rename in git.
4. We now handle ActivityContentProfile announcements from the chain. These get stored in the profileSlice in redux. I've changed the structure of the profile slice to be a map keyed by dsnpId (aka socialAddress).
5. In `PostList.txs` we now match FeedItems to their associated Profiles and use that to substitute the user's name for their dsnpId. This is a bit of a proof-of-concept that we can associate this data. The actual UX may need to change.
6. In the process of fixing the tests for this PR, I addressed multiple warnings that were unrelated, but polluting the test output.

Steps to Verify
----------------
1. Start with a clean chain with deployed contracts.
2. run `node scripts/populate.mjs  --experimental-vm-modules` to create some content.
3. `npm run start`.
4. Click `All posts`.
5. You should see ~20 posts with names instead of ids.
6. `npm run test`
7. Verify test output is clean.


package.json
```diff
@@ -201,12 +202,14 @@
...
     "less-loader": "^6.2.0",
     "michelangelo": "^0.8.0",
+    "node-fetch": "^2.6.1",
     "optimize-css-assets-webpack-plugin": "5.0.4",
     "pnp-webpack-plugin": "1.6.4",
     "prettier": "2.2.1",
     "react-dev-utils": "^11.0.4",
     "react-test-renderer": "17.0.1",
     "redux-devtools-extension": "^2.13.9",
+    "request": "^2.88.2",
```
enddynayn, line 212: is this something that we need?

acruikshank: Nope. Good catch.

src/redux/slices/profileSlice.ts
```diff
@@ -2,33 +2,28 @@ import { createSlice, PayloadAction } from "@reduxjs/toolkit";
...
+      state.profiles = { ...state.profiles, [key]: newProfile };
       return state;
     },
     removeProfile: (state, action: PayloadAction<HexString>) => {
-      const oldProfile = state.profiles.find(
-        (profile) => profile.socialAddress === action.payload
-      );
-      if (oldProfile)
-        state.profiles.splice(state.profiles.indexOf(oldProfile), 1);
+      delete state.profiles[action.payload];
```
enddynayn, line 26: I do not think we want to mutate the current state but instead return a new one. Another way is using filter.

src/utilities/activityContent.ts
```diff
@@ -0,0 +1,63 @@
...
+  uriList: string[]
+): ActivityContentNote => {
+  const activityContent: ActivityContentNote = {
+    "@context": "https://www.w3.org/ns/activitystreams",
+    type: "Note",
+    content: note,
+    published: new Date().getTime().toString(16),
+  };
+  if (uriList) {
+    if (typeof uriList === "string") {
```
enddynayn, line 20: uriList string is an array of strings how could this happen?

src/utilities/activityContent.ts
```diff
@@ -0,0 +1,63 @@
+import path from "path";
```
enddynayn, line 1: path?

src/utilities/activityContent.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (uriList) {
+    if (typeof uriList === "string") {
+      uriList = [uriList];
+    }
+    activityContent["attachment"] = uriList.map(
+      (item): ActivityContentAttachment => {
+        const extension = path.extname(item).replace(".", "");
+        const supportedVideoDomains = /youtu.be|youtube.com|vimeo.com/;
+        if (supportedVideoDomains.test(item)) {
+          return createVideo(item, "text/html", { algorithm: "" }, 0, 0);
```
enddynayn, line 28: maybe the logic for the type of video could be extracted to its own function

enddynayn (commented):
Left a couple comments

src/redux/slices/profileSlice.ts
acruikshank: Yeah, I started to write an immutable version. but noticed that the old version was using splice. I'll fix it.

src/utilities/activityContent.ts
acruikshank: Yep. This probably became unnecessary in the translation.

src/services/wallets/torus/torus.ts
```diff
@@ -57,6 +57,7 @@ export const getPublicAddress = async (
 
 export const getWalletAddress = async (): Promise<HexString> => {
   const { verifier, verifierId } = await getUserInfo();
+  console.log("WALLET ADDR: ", verifier, verifierId);
```
claireolmstead, line 60: console.log

src/utilities/activityContent.ts
acruikshank: This was already in the file (this change is mostly a filename that git misinterpreted). I think webpack is polyfiling this node package.

src/utilities/activityContent.ts
acruikshank: good call. extracted.

src/services/wallets/torus/torus.ts
acruikshank: good catch. removed.

src/redux/slices/profileSlice.ts
acruikshank: fixed.

enddynayn (approved):

PR #38, opened 2021-07-27
WIP Feat/remove unnecessary features #179011705

shannonwells:
Purpose
---------------
Remove features that will not be used for the demo.
[Finishes #179011705](https://www.pivotaltracker.com/story/show/179011705)

Change summary
---------------
* Remove NewPost, NewPostImageUpload components and styles
* Hard code handle, name, and image URL (which may change)
* Disable inputs for Profile component
* Remove style guide

Steps to Verify
----------------
You should be able to build and run the app and click around, viewing the Feed without errors.

----------------


acruikshank:
Is this PR in the right repo? We don't want to remove these things from the example-client.

PR #40, opened 2021-07-31
Chores/update to sdk 2.0.0 & fix breakages

shannonwells:
## Purpose
The big SDK refactor causes a bunch of breakage in example-client if you update to latest. Fix these.
[Fixes #179076239](https://www.pivotaltracker.com/story/show/179076239)

## Solution
1. Upgrade to **SDK 2.0.0** and fix breakages
2. update a couple of packages to fix some vulnerabilities.
3. Get rid of a couple unused packages
4. Update to next nodejs minor version

## To Verify
- All tests should pass
- You should be able to run the example-client, login and create new posts without any errors -- that is, assuming you hold your mouth right, it's a first Tuesday in the month  with a waxing gibbous moon in the 3rd quarter and you've done all the static-server and chain setup correctly.


src/components/PostMedia.tsx
```diff
@@ -13,39 +13,32 @@ interface PostMediaProps {
...
-        },
-        index: number
-      ) => (
+    return attachment?.map((item, index) => {
+      return (
         <div key={index} className="PostMedia__cover">
           {item.type.toLowerCase() === "image" && (
             <a
-              href={item.url as string}
+              href={item.url[0].href}
```
shannonwells, line 21: I'm sure there needs to be some selection process here, but I think that's intended to be left as an exercise for the developer.

src/services/sdk.ts
```diff
@@ -125,7 +124,7 @@ const buildPublication = (batchData: BatchFileData): Publication => {
 const dispatchActivityContent = (
   dispatch: Dispatch,
   message: BroadcastAnnouncement,
-  activityContent: ActivityContent,
+  activityContent: ActivityContentNote | ActivityContentProfile,
```
shannonwells, line 127: I'm not sure what type is expected; this is just to make it work, as ActivityContent went away.

src/services/sdk.ts
```diff
@@ -181,7 +184,7 @@ const dispatchProfile = (
 
   dispatch(
     upsertProfile({
-      ...profile.describes,
```
shannonwells, line 184: Not sure what this was supposed to be.

src/components/PostMedia.tsx
rosalinekarr: yeah, ideally we should scan over each link and choose the one with...

1) the smallest resolution larger than size we intend to display, and
2) the highest quality format supported in the current browser

but I think we can save that for a future PR

src/services/sdk.ts
rosalinekarr: This is the correct type. We took out `ActivityContent` because we wanted to be more explicit about what was profile data and what was post data.

src/services/sdk.ts
```diff
@@ -159,7 +158,7 @@ const buildPublication = (
 const dispatchActivityContent = (
   dispatch: Dispatch,
   message: BroadcastAnnouncement,
-  activityContent: ActivityContent,
+  activityContent: ActivityContentNote | ActivityContentProfile,
   blockNumber: number
 ) => {
   switch (activityContent.type) {
```
rosalinekarr, line 164: Non-blocking: the [`isActivityContentNote()`](https://github.com/LibertyDSNP/sdk-ts/blob/main/src/core/activityContent/validation.ts#L173) and [`isActivityContentProfile()`](https://github.com/LibertyDSNP/sdk-ts/blob/main/src/core/activityContent/validation.ts#L200) methods in `sdk/core/activityContent` might save you the trouble of casting the type on lines 169 and 176 below.

src/services/sdk.ts
```diff
@@ -187,15 +186,21 @@ const dispatchFeedItem = (
...
+
   dispatch(
     addFeedItem({
       fromAddress: decoder.decode((message.fromId as any) as Uint8Array),
       blockNumber: blockNumber,
       hash: decoder.decode((message.contentHash as any) as Uint8Array),
-      timestamp: new Date(content.published).getTime(), // TODO: really wrong
+      timestamp: timestamp,
       uri: decoder.decode((message.url as any) as Uint8Array),
       content: {
```
rosalinekarr, line 201: Non-blocking: there's a [`createNote()` method](https://github.com/LibertyDSNP/sdk-ts/blob/main/src/core/activityContent/factories.ts#L191) in `sdk/core/activityContent` that could simplify this for you.

src/services/sdk.ts
rosalinekarr: Previously, `"Profile"` activity content objects had a nested `"Person"` object under the describe key with the username on it, but they're the same thing now.

src/test/testFeeds.ts
```diff
@@ -9,73 +9,27 @@ import {
...
  * @param handle the new username of the profile update
  */
-export const generatePerson = (name?: string): ActivityContentPerson => {
+export const generatePerson = (name?: string): ActivityContentProfile => {
   return {
     "@context": "https://www.w3.org/ns/activitystreams",
     name: name || "",
-    type: "Person",
+    type: "Profile",
     published: "17ad9cb8551",
```
rosalinekarr, line 33: Non-blocking: `published` isn't actually required on the ActivityContent anymore now that we added it to the announcements.

src/test/testFeeds.ts
```diff
@@ -116,6 +70,13 @@ export const getPrefabFeed = (): FeedItem<ActivityContentNote>[] => {
   const address1 = getPrefabSocialAddress(1);
   const address2 = getPrefabSocialAddress(2);
   const address3 = getPrefabSocialAddress(3);
+
+  const imgAttachment = generateImageAttachment(
+    "https://64.media.tumblr.com/bd8d2127a91f57463c2e753cf837ab6e/014df86f4004efef-ec/s1280x1920/adda1023806b71606f83f484a64daa03bce12c8d.jpg"
+  );
+  const note = generateNote("Everyone leave me alone");
+  note.attachment = [imgAttachment];
```
rosalinekarr, line 78: [`generateNote()`](https://github.com/LibertyDSNP/sdk-ts/blob/main/src/generators/activityContentGenerators.ts#L24) has an optional `hasAttachment` argument which will automatically add a fake attachment for you.

src/test/testFeeds.ts
```diff
@@ -150,13 +103,15 @@ export const getPrefabFeed = (): FeedItem<ActivityContentNote>[] => {
 export const generateRandomNote = (): ActivityContentNote => {
   const message = getRandomMessage();
   const attachment = getRandomAttachment();
-  return generateNote(message, attachment);
+  const note = generateNote(message);
```
rosalinekarr, line 106: Same thing here

src/utilities/activityContent.ts
```diff
@@ -1,26 +1,41 @@
...
+  createVideoAttachment,
+  ActivityContentHash,
+  createImageAttachment,
+  createImageLink,
 } from "@dsnp/sdk/core/activityContent";
 
 const createMediaAttachment = (item: string): ActivityContentAttachment => {
   const extension = path.extname(item).replace(".", "");
   const supportedVideoDomains = /youtu.be|youtube.com|vimeo.com/;
+  const activityContentHashes: Array<ActivityContentHash> = [{ algorithm: "" }];
```
rosalinekarr, line 15: This will actually fail validation if there isn't at least one valid keccak256 content hash. There's a [`createHash` helper](https://github.com/LibertyDSNP/sdk-ts/blob/main/src/core/activityContent/factories.ts#L395) in `sdk/core/activityContent` which takes the content to hash as a string and spits out a properly formatted hash object.

rosalinekarr (changes requested):
One blocking note regarding content hashes, but otherwise looks good

src/test/testFeeds.ts
shannonwells: oh! I missed that completely, ok.

src/utilities/activityContent.ts
shannonwells: This is what was being used throughout this file, so I guess the validation changed?

rosalinekarr (approved):
Looks like GitHub recommends deleting `getRandomAttachment`. Other than that, this LGTM

src/components/PostMedia.tsx
```diff
@@ -2,50 +2,40 @@ import React from "react";
...
```
claireolmstead, line 18: will the url always be saved as an array? that seems like overkill if each attachment in each item in the map has an array of length 1.  why not save it as a string?

src/components/PostList.tsx
```diff
@@ -55,11 +55,13 @@ const PostList = ({ feedType }: PostListProps): JSX.Element => {
             .slice(0)
             .reverse()
             .map((post, index) => {
-              const namedPost = {
+              const fromAddress = profiles[post.fromAddress]
+                ? profiles[post.fromAddress].name || ""
```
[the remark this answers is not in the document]
shannonwells: This was only for type conversion.  TypeScript was complaining about setting "string or undefined" to a string.  I wasn't sure if I should set up an error if either of these was undefined.

src/components/PostMedia.tsx
shannonwells: it's to do with the attachment actually being potentially different sizes or codecs.  See @rosalinekarr 's comment.  For now we're just going to always use the first one and not bother checking whether there are others there and if we should use them.

src/components/PostList.tsx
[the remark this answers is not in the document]
shannonwells: I made it more strict.

PR #51, opened 2021-08-12
Feat/record and display handles

acruikshank:
Purpose
---------------
We need to show handle information in the example client. Unfortunately, this information isn't available from the profiles we are already retrieving, so it must be derived from DSNP registry update events.

Solution
---------------
This PR adds a subscription for DSNP registry events. The event data is used to associate users' handles with their profile data. This information is stored in the profile slice of the redux store. We use it to show the handle in the current user's profile and in the posts feed if the sender's name is not available.

Change summary
---------------
* Update to latest SDK (to get registry update subscription feature)
* Modify populate script to add handles to accounts and register them (note there are some fixes here to solve some concurrency issues).
* Change user slice to store DSNP id rather than profile. Th profile is then associated with a profile from profile slice. At login, the current user DSNP id is currently hard coded to one of the id's generated in the populate script to demonstrate this association. This will be replaced by an address to DSNP id lookup in a future story.
* Show handle as well as name in PostList.
* Show handle in current user Profile.

Steps to Verify
----------------
1. Start with a clean chain with contracts deployed.
2. Run populate script `node scripts/populate.mjs`
3. Start example-client.
4. Log in with metamask.
5. Note that Profile now has name, handle and social address (DSNP id) of one of the accounts in the populate script (probably "An Bai").
6. Click on "All Posts".
7. Note that all feed items have a name and handle.


package.json
```diff
@@ -7,7 +7,7 @@
   "dependencies": {
     "@dsnp/contracts": "^0.0.0-5c945f",
     "@dsnp/parquetjs": "^1.1.0",
-    "@dsnp/sdk": "2.0.0",
+    "@dsnp/sdk": "^0.0.0-6082ae",
```
claireolmstead, line 10: why roll back?

src/components/FromLine.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Profile } from "../utilities/types";
+
+interface FromLineProps {
+  profile: Profile;
+}
+
+export const FromLine = ({ profile }: FromLineProps): JSX.Element => {
```
claireolmstead, line 8: change the name to `FromTitle`

src/components/PostList.tsx
```diff
@@ -54,19 +47,9 @@ const PostList = ({ feedType }: PostListProps): JSX.Element => {
...
             .slice(0)
             .reverse()
-            .map((post, index) => {
-              if (!post.fromAddress)
-                throw new Error(`no fromAddress in post: ${post}`);
-              const fromAddress: string = profiles[post.fromAddress]
-                ? (profiles[post.fromAddress].name as string)
-                : post.fromAddress;
-
-              const namedPost: FeedItem<ActivityContentNote> = {
```
claireolmstead, line 64: 👍🏼

src/components/Profile.tsx
```diff
@@ -4,13 +4,21 @@ import ConnectionsList from "./ConnectionsList";
...
+  const profiles: Record<types.HexString, types.Profile> = useAppSelector(
+    (state) => state.profiles?.profiles || {}
+  );
+
+  const profile: types.Profile | undefined = userId
+    ? profiles[userId]
+    : undefined;
 
-  const handle = "insert_handle_here";
+  const handle = profile?.handle;
```
claireolmstead, line 21: woohoo!

src/redux/slices/userSlice.ts
```diff
@@ -1,37 +1,32 @@
...
+import { DSNPUserId } from "@dsnp/sdk/dist/types/core/identifiers";
 
 interface UserState {
-  profile?: Profile;
+  id?: DSNPUserId;
   walletType: wallet.WalletType;
 }
 
 const initialState: UserState = {
-  profile: session.hasSession() ? session.loadSession()?.profile : undefined,
```
claireolmstead, line 12: does this affect the session storage?

claireolmstead (approved):

package.json
acruikshank: This is actually moving forward. The method to subscribe to registry events is not yet in a release.

src/redux/slices/userSlice.ts
acruikshank: It does. Good catch. I added it back using id instead.

PR #52, opened 2021-08-12
Feature/convert types/179171825

claireolmstead:
Purpose
---------------
[https://www.pivotaltracker.com/story/show/177349627](url)

Change types to match Spec.

Solution
---------------
* use types imported from the sdk to define Profile and FeedItem and any other types that need to be changed.

Change summary
---------------
* changed naming
* changed type
* adjusted component to display new type correctly
* updated tests

Steps to Verify
----------------
1. create a new post from the button
2. DO NOT create a new post from the script

----------------
* Types changed on the populate script but name/fromId is not displaying yet.  In a different story


rosalinekarr (approved):
LGTM, just two non-blocking suggestions

PR #58, opened 2021-08-17
Retrieve Graph Data from DSNP

acruikshank:
Purpose
---------------
In order to know who a user follows (or unfollows), we need to retrieve graph change event from DSNP batch data.

Solution
---------------
This PR retrieves all graph change events from DSNP batches and uses it to populate the graph in Redux. It then uses that to update the UI in various places like the connection list, and the user's feed.

Change summary
---------------
* Alter populate script to add follow information
* Update the SDK to incorporate a parquet issue storing graph change events
* Read graph change events in the SDK
* Alter the redux store to store follows and followers efficiently
* Modify PostLIst to use the new graph change events.
* Modify ConnectionList to use the new graph change events.
* Fix tests.
* Update package-lock.json to version 2 (hopefully for the last time).

Steps to Verify
----------------
1. Start up a clean, deployed chain and the static server.
2. Run the new populate script.
3. Start the app.
4. Confirm that there are posts in the "Feed" tab of the post list.
5. Confirm that you see connections when you click "followers" in the profile block.
6. Confirm that you see connections when you click "following" in the profile block.
7. Confirm that the posts in the feed are consistent with the following list.
8. Confirm that the the follow/unfollow buttons in the "followers" section are consistent with the follow list.

PR #61, opened 2021-08-17
save profiles

acruikshank:
Purpose
---------------
When we edit a profile, we want to actually send the change out to the DSNP network so others know about it and the change is persisted.

Solution
---------------
This PR wires the edit profile functionality to a new SDK method that creates and publishes a new profile.

Change summary
---------------
* Create `sdk.saveProfile` method (with some sdk refactoring to reduce redundancy).
* Wire the Profile edit functionality to call the save profile.
* Track blockNumber, blockIndex and batchIndex, so we can correctly order profile changes.
* Add comparison logic in the profile slice of the redux store to ensure we always show the latest profile.
* Address some feedback that came in late from #58.

Steps to Verify
----------------
1. Set up the app with the usual steps.
2. Start the app.
3. Log in.
4. Hit the edit button in the profile section.
5. Change the name.
6. Expect to sign a transaction.
7. Wait for the transaction to clear.
8. Expect every instance of your name in the feeds to be updated to the new value.
9. Reload the page.
10. Expect the name to be updated in the profile section.

### DIFF
--- a/src/components/Profile.tsx
+++ b/src/components/Profile.tsx
@@ -46,14 +46,13 @@ const Profile = (): JSX.Element => {
   };
 
   const saveEditProfile = async () => {
-    if (userId !== undefined && newName !== undefined) {
-      const newProfile = core.activityContent.createProfile({
-        name: newName,
-        icon: profile?.icon,
-      });
-      await saveProfile(userId, newProfile);
-    }
     setIsEditing(!isEditing);
+    if (userId === undefined || newName === undefined) return;
+    const newProfile = core.activityContent.createProfile({
+      name: newName,
+      icon: profile?.icon,
+    });
+    await saveProfile(userId, newProfile);
   };
 
   const cancelEditProfile = () => {
--- a/src/components/Profile.tsx
+++ b/src/components/Profile.tsx
@@ -5,6 +5,8 @@ import React, { useEffect, useState } from "react";
 import { useAppSelector } from "../redux/hooks";
 import * as types from "../utilities/types";
 import { DSNPUserId } from "@dsnp/sdk/dist/types/core/identifiers";
+import { saveProfile } from "../services/sdk";
+import { core } from "@dsnp/sdk";
 
 const Profile = (): JSX.Element => {
   const userId: DSNPUserId | undefined = useAppSelector(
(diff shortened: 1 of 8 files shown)


src/redux/slices/profileSlice.ts
```diff
@@ -9,6 +9,16 @@ const initialState: profileState = {
   profiles: {},
 };
 
+const latestProfile = (existing: Profile, current: Profile): Profile => {
+  if (!existing || existing.blockNumber === undefined) return current;
```
enddynayn, line 13: what the diff between existing vs current?

src/components/Profile.tsx
```diff
@@ -43,15 +45,21 @@ const Profile = (): JSX.Element => {
...
-  const saveEditProfile = () => {
-    //this is where we write to the blockchain
+  const saveEditProfile = async () => {
+    if (userId !== undefined && newName !== undefined) {
+      const newProfile = core.activityContent.createProfile({
+        name: newName,
+        icon: profile?.icon,
+      });
+      await saveProfile(userId, newProfile);
+    }
```
enddynayn, line 55: do we need to wait util we get a response back before setting `setIsEditing(!isEditing)`?

src/components/Profile.tsx
```diff
@@ -43,15 +45,21 @@ const Profile = (): JSX.Element => {
       : `ProfileBlock__${sectionName}`;
   };
 
-  const saveEditProfile = () => {
-    //this is where we write to the blockchain
+  const saveEditProfile = async () => {
+    if (userId !== undefined && newName !== undefined) {
```
enddynayn, line 49: maybe return early?
if (userId == undefined || newName == undefined)  return;

src/services/sdk.ts
```diff
@@ -106,8 +117,18 @@ export const startPostSubscription = (
...
 
   // subscribe to all announcements
+  let blockNumber: number;
+  let blockIndex = 0;
   core.contracts.subscription.subscribeToBatchPublications(
-    handleBatchAnnouncement(dispatch),
+    (announcement: BatchPublicationLogData) => {
+      if (announcement.blockNumber !== blockNumber) {
+        blockNumber = announcement.blockNumber;
+        blockIndex = 0;
```
enddynayn, line 126: what is block index?

src/components/Profile.tsx
acruikshank: not really.

src/components/Profile.tsx
acruikshank: ok.

enddynayn (approved):

src/services/sdk.ts
acruikshank: blockIndex is order this batch announcement is received within the block. Announcements found in batches that are published earlier in the block are considered to have come earlier according to the spec.

PR #62, opened 2021-08-18
Features/create identity registration #178933068

shannonwells:
Purpose
---------------
Functionality only, no styling, creates a registration on first time logging into DSNP network via example client. 
[Part of #178933068](https://www.pivotaltracker.com/story/show/178933068)

Solution
---------------
### In:
The first time logging into the example client, asks you to create a new handle and then finishes up the login process.

### Out:
* Styling to match design
* Reconciliation with choice of handles on first login if registration(s) already exist
* Checking if handle exists. This will be in a separate PR.

with @acruikshank , we will follow up in his PR by merging components and doing styling.

Change summary
---------------
- [x] New "Register" component that has a form for creating a new handle, and calls sdk.createRegistration with choice of handle and returns new DSNP userId to the Login component so it can finish up login and update user and profile state.
- [x] Basic UI w/o real styling 
- [x] if registration fails, unset everything so user can start over 
- [x] Tests 
- [x] CI passing 

Steps to Verify
----------------
1.  Do setup of local hardhat node + static server + npm run start on this branch
1.  Login with a new account through Metamask that has no existing registration
1.  See the create handle form popup, enter a handle and submit
1.  Verify that the button reads Log out in the top right
1.  Verify that your new social ID, new handle and icon appear
1.  If you've run the populate script, verify that you can see posts in All Posts.

----------------
Screen Shot 2021-08-18 at 11 44 47 AM
giphy1

### DIFF
--- a/scripts/populate.mjs
+++ b/scripts/populate.mjs
@@ -45,10 +45,6 @@ const accounts = [
       ),
     ],
     follows: [4, 13, 14, 18],
-    tag: [
-      {name: "cereal"},
-      {name: "salad"},
-    ]
   },
   {
     address: "0x70997970c51812dc3a010c7d01b50e0d17dc79c8",
@@ -62,11 +58,6 @@ const accounts = [
     ],
     text: "Hot take: Hotdogs aren't a sandwich, they're a taco",
     follows: [7, 15, 17, 19],
-    tag: [
-      {name: "#taco"},
-      {name: "#hotdogs"},
-      {name: "this is another hashtag"}
-    ]
   },
   {
     address: "0x3c44cdddb6a900fa2b585dd299e03d12fa4293bc",
@@ -83,7 +74,6 @@ const accounts = [
       "see the light at the end of the tunnel, but if you just " +
       "keep moving, you will come to a better place.",
     follows: [9, 13, 14, 15, 19],
-    tag: {name: "a single tag"},
   },
   {
     address: "0x90f79bf6eb2c4f870365e785982e1f101e93b906",
@@ -97,7 +87,6 @@ const accounts = [
     ],
     text: "Сухие завтраки - это разновидность салата. Конец истории.",
     follows: [5, 15, 17],
-    tag: ["cereal","salad","cухие","салата"]
   },
   {
     address: "0x15d34aaf54267db7d7c367839aaf71a00a2c6a65",
@@ -109,7 +98,6 @@ const accounts = [
     text: "This Vimeo -- WAT. Amirite?",
     name: "Louis Bollen",
     follows: [11, 13, 16],
-    tag: [{name: "not foodblogger"},{name: "shouldn't appear"}]
   },
   {
     address: "0x9965507d1a55bcc2695c58ba16fb37d819b0a4dc",
@@ -119,7 +107,6 @@ const accounts = [
     text:
       "My favorite sea shanty is 'Friggin in the Riggin'. I don't like attachments. That's why I don't have any.",
     follows: [11, 14, 15, 16, 17],
-    tag: {name: "another single tag"},
   },
   {
     address: "0x976ea74026e726554db657fa54763abd0c3a0aa9",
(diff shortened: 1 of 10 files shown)


acruikshank (approved):
This looks good. There's a lot of overlap with #65, but I think we can deal with that.

src/components/Login.tsx
```diff
@@ -16,24 +17,48 @@ const Login = ({ loginWalletOptions }: LoginProps): JSX.Element => {
...
+    loginWalletOptions || wallet.WalletType.NONE
+  );
 
   const dispatch = useAppDispatch();
   const userId = useAppSelector((state) => state.user.id);
-  const walletType = useAppSelector((state) => state.user.walletType);
+
+  const setLoginAndSession = (fromId: string) => {
+    setRegistrationVisible(false);
+    dispatch(userLogin({ id: fromId, walletType }));
```
enddynayn, line 34: Im curious to learn why we don't also store the walletAddress in Redux?

src/components/Login.tsx
```diff
@@ -16,24 +17,46 @@ const Login = ({ loginWalletOptions }: LoginProps): JSX.Element => {
...
+    session.saveSession({ id: undefined, walletType });
+  };
 
   const login = async (walletType: wallet.WalletType) => {
     if (loading) return;
     startLoading(true);
+    setWalletType(walletType);
     try {
-      const walletAddress = await wallet.wallet(walletType).login();
+      const walletAddrLocal = await wallet.wallet(walletType).login();
```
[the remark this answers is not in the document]
shannonwells: it's a convention I adopt sometimes to note that something is a local variable as distinguished from another one with a similar name.  It was confusing me to leave it.

scripts/populate.mjs
```diff
@@ -491,6 +513,7 @@ for await (let account of accounts.values()) {
     { attachment: account.attachment }
   );
   content.published = new Date().toISOString();
+  if (account.tag) content.tag = account.tag
```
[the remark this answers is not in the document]
shannonwells: I meant actually to take it out. I had put it there to test some Foodee stuff.

src/components/Login.tsx
```diff
@@ -16,24 +17,48 @@ const Login = ({ loginWalletOptions }: LoginProps): JSX.Element => {
...
-      const fromId = await sdk.getSocialIdentity(walletAddress);
+      const fromId = await sdk.getSocialIdentity(walletAddrLocal);
       if (fromId) {
-        dispatch(userLogin({ id: fromId, walletType }));
-        session.saveSession({ id: fromId, walletType });
+        setLoginAndSession(fromId);
+      } else {
+        setWalletAddress(walletAddrLocal);
+        setRegistrationVisible(true);
       }
```
enddynayn, line 59: this can be removed from setLoginAndSession and added after the else setRegistrationVisible(true);

src/components/Register.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const register = async (formValues: { handle: string }) => {
+    try {
+      const userId = await createRegistration(
+        registerOptions.walletAddress,
+        formValues.handle
+      );
+      if (userId) {
+        registerOptions.onSuccess(userId.replace("dsnp://", ""));
+      }
+      setCreateHandleFormVisible(false);
```
enddynayn, line 26: could we remove could we remove  setCreateHandleFormVisible(false); since we have it in finally?

scripts/populate.mjs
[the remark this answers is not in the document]
acruikshank: I think it's better to keep it. We're going to want it at some point.

enddynayn (approved):

src/components/Login.tsx
shannonwells: Not sure, perhaps we don't need it once we login and maybe it's a security issue.

scripts/populate.mjs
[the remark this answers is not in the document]
shannonwells: I'm going to put it back in a small PR.

src/components/Login.tsx
acruikshank: Mainly because we don't need it. It gets stored in the `signer` in the sdk config so that's how it's used in signing. Everything else is keyed off DSNP id.

PR #65, opened 2021-08-19
pick handle at login

acruikshank:
Purpose
---------------
It's possible for the same wallet address to be authorized as a delegate to multiple DSNP ids, so logging in to a wallet is not necessarily sufficient to determine which account to display in the example client.

Solution
---------------
This PR adds a modal to disambiguate the DSNP id and handle when there are multiple registrations for a wallet address. We were already retrieving all the registrations. This PR is just adds UI to pick between them rather than automatically picking the first one.

Change summary
---------------
* Add a second registration to the populate script for validation (and refactor a bit to permit accounts without posts).
* Modify the Login component to hold off on login and show the new RegistrationModal when multiple registrations are received.
* Implement registration modal to select registration.
* Style registration modal.

Steps to Verify
----------------
1. Set up app (hardhat, deploy, static server) and run populate.
2. Start app and log in with metamask using the `0x8626f6940e2eb28930efb4cef49b2d1f2c9c1199` account.
3. Expect a modal to appear showing two registrations.
4. Select one of the registrations and click "Select Handle".
5. Expect the app to be logged in and the account to match the selected registration.
6. Log out and log in again with the other registration.
7. Expect the account to be changed.

----------------
- Screen Shot 2021-08-19 at 3 08 59 PM


src/test/testAddresses.ts
```diff
@@ -21,7 +21,7 @@ export const generateDsnpUserId = (): HexString => {
  */
 export const getPrefabWalletAddress = (index: number): HexString => {
   const regex = /0/gi;
-  const address = "0x" + "CODE0000".replace(regex, index.toString());
+  const address = "0x" + "CADE0000".replace(regex, index.toString());
```
shannonwells, line 24: why did these change?

shannonwells (changes requested):
We'll have to pair up and figure out how best to combine these two.

src/test/testAddresses.ts
acruikshank: They fail validation. O is not a valid hexadecimal character.

shannonwells: Ah. I thought it was a zero. #fonts

PR #74, opened 2021-08-24
fix feed loading

acruikshank:
Purpose
---------------
A bug was introduced into the feed reducer that prevented all existing feed items from entering redux state and displaying in the feed. This PR attempts to keep the spirit of the erroneous code with limiting the posts.

Solution
---------------
A condition was added in the feed reducer that prevented the item from being added to the store if it did not match other state set by the NewPost component. This would reject all posts from other people in order to avoid the wrong posts changing the loading state. This PR limits this condition only to the loading state and allows all posts into state.

Change summary
---------------
* Remove `(state.isPostLoading.currentUserId === newFeedItem.fromId)` clause.
* Replace with `newPostLoadingState` and `newReplyLoadingState`

Steps to Verify
----------------
1. Start up app with a populated feed.
2. Existing posts appear

### DIFF
--- a/src/redux/slices/feedSlice.ts
+++ b/src/redux/slices/feedSlice.ts
@@ -23,22 +23,36 @@ const initialState: feedState = {
   isReplyLoading: { loading: false, parent: undefined },
 };
 
+const newPostLoadingState = (
+  state: isPostLoadingType,
+  item: FeedItem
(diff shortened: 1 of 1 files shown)


shannonwells (approved):
Good enough 4n

PR #79, opened 2021-08-27
Feature/show others profiles/177348273

claireolmstead:
Purpose
---------------
Feature: As a user, I want to see other users' profiles.
[https://www.pivotaltracker.com/story/show/177348273](url)

Solution
---------------
when you click on another user's profile picture from the feed, I want the profile component to load in that user's profile information
acceptance criteria:

switches from your profile information to the selected user's profile information
back arrow navigates back to your profile
does NOT show other user's corresponding feed posts
does NOT include follow/unfollow functionality

Change summary
---------------
* Show other users' profiles on click of their avatar or name/handle by adding a reducer to the user redux store
* make profile block stick to top
* remove edit button if not your profile
* create back button to go back to your profile
* create header in nav to show who's profile you're seeing
* write tests

Steps to Verify
----------------
1. login
2. click on other users' profiles
3. logout
4. log back in and see that it shows your profile
5. click on another users' profile
6. press the back button to see your profile come back

----------------
https://user-images.githubusercontent.com/43625033/131540560-ddd31d1e-5dea-4f0f-a8bb-b03b52daf47d.mov


src/components/Feed.tsx
```diff
@@ -2,16 +2,32 @@ import React, { useState } from "react";
...
+import { useAppDispatch, useAppSelector } from "../redux/hooks";
+import * as types from "../utilities/types";
+import { ArrowLeftOutlined } from "@ant-design/icons";
+import { setDisplayedProfileId } from "../redux/slices/userSlice";
 
 enum FeedTypes {
   MY_FEED,
   MY_POSTS,
   DISCOVER,
+  PROFILE_POSTS,
```
claireolmstead, line 14: not being fully used yet, but will come in in this story and wanted the onclick to do nothing here [https://www.pivotaltracker.com/story/show/177349009](url)

src/components/Feed.tsx
```diff
@@ -2,16 +2,32 @@ import React, { useState } from "react";
...
 }
 
 const Feed = (): JSX.Element => {
+  const dispatch = useAppDispatch();
+
   const userId = useAppSelector((state) => state.user.id);
+  const currentProfileId = useAppSelector(
```
claireolmstead, line 21: I want to figure out the best naming for this and use it throughout. if user.id = userId, user.displayedProfileId = ?

src/components/Login.tsx
```diff
@@ -71,7 +71,7 @@ const Login = ({ loginWalletOptions }: LoginProps): JSX.Element => {
 
   useEffect(() => {
     ethereum?.on("accountsChanged", () => logout());
-  }, []);
+  }, [logout]);
```
acruikshank, line 74: Blocking: I'm not sure what prompted this, but it isn't right. Logout gets defined every render, so this method will get called constantly, and we'll keep redefining the subscription. That's a guaranteed memory leak. I'm not sure when the ethereum subscription got added, but this isn't the right place for that either.

src/components/Profile.tsx
```diff
@@ -9,9 +9,12 @@ import { saveProfile } from "../services/sdk";
 import { core } from "@dsnp/sdk";
 
 const Profile = (): JSX.Element => {
+  const myId: DSNPUserId | undefined = useAppSelector((state) => state.user.id);
```
acruikshank, line 12: It'd be better to keep this one as `userId` an make the other `displayId`, `idToDisplay` or some such, since we use `userId` throughout the app to mean the current logged in user.

src/redux/slices/userSlice.ts
```diff
@@ -6,11 +6,15 @@ import { DSNPUserId } from "@dsnp/sdk/dist/types/core/identifiers";
 interface UserState {
   id?: DSNPUserId;
   walletType: wallet.WalletType;
+  displayedProfileId?: DSNPUserId;
 }
 
 const initialState: UserState = {
   id: session.hasSession() ? session.loadSession()?.id : undefined,
   walletType: session.loadSession()?.walletType ?? wallet.WalletType.NONE,
+  displayedProfileId: session.hasSession()
```
acruikshank, line 15: This shouldn't be necessary. An undefined `displayedProfileId` should behave the same as when the `displayedProfileId === id`.

src/redux/slices/userSlice.ts
```diff
@@ -21,10 +25,14 @@ export const userSlice = createSlice({
       ...state,
       ...action.payload,
     }),
-    userLogout: (_state) => ({ walletType: wallet.WalletType.NONE }),
+    userLogout: (_state) => ({
+      walletType: wallet.WalletType.NONE,
+      displayedProfileId: undefined,
```
acruikshank, line 30: If this were right, this line would be redundant, but I don't think this is right. It should be possible to click on a profile from the the `Discover` feed even when you're not logged in, so it doesn't really make sense for logging out to affect the `displayedProfileId` state. If this is the correct behavior, this line should be `displayedProfileId: state.displayedProfileId`.

acruikshank (changes requested):

acruikshank (approved):
Looks great!

PR #80, opened 2021-08-31
Feature/annoucement id copy button/179257443

claireolmstead:
Purpose
---------------
copy announcement URI instead of just the hash

Solution
---------------
create announcement URI based off of the spec: [https://spec.dsnp.org/Identifiers#dsnp-announcement-uri](url)

Change summary
---------------
* bring in fromId to PostHashDropdown
* create proper announcementURI

Steps to Verify
----------------
1. Check that value copied is correct

### DIFF
--- a/src/components/Post.tsx
+++ b/src/components/Post.tsx
@@ -63,7 +63,7 @@ const Post = ({ feedItem }: PostProps): JSX.Element => {
           }
         />
       </div>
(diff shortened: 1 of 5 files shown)


acruikshank (approved):
This looks great. It's going to need to change with the SDK 3.0, but that's ok.

PR #82, opened 2021-09-01
add ability to follow and unfollow users

acruikshank:
Purpose
---------------
Users need a way to follow and unfollow each other.
https://www.pivotaltracker.com/story/show/177348398

Solution
---------------
This PR adds the ability to follow and unfollow users that are following the current user and the ability to unfollow users the logged in user is currently following. It lays the grounds work for following arbitrary users, but we are waiting for #79 for the UI to implement that functionality.

Change summary
---------------
* Add sdk methods to follow and unfollow.
* Modify ConnectionListProfiles to call those methods when the existing buttons are pressed.
* Modify the graphSlice in redux to store a RelationshipStatus enum instead of a boolean, so we can track when a following/followed by relationship has been updated by the user, but not yet recorded on chain.
* Add styling to the graph relationship buttons for the updating state.

Steps to Verify
----------------
1. Start the app in a clean populated state.
2. Open the "followers" tab in the profile block
3. Click one of the "follow" buttons for a user that is not yet followed and approve the transaction.
4. Note that the button you clicked is now in the "updating state.
5. Open the "following" tab and note that the new follow has been added (probably still in the updating state).
6. Note that you see posts from the new follow in the "My Feed" section.
7. Wait for the transaction to go through.
8. Note that the updating button is replaced with an "unfollow" button in the "Followers" and "Following" tabs. 
9. Reload the page, and note that all the changes above still exist once the page is fully loaded.
10. In the "following" tab, choose a user that is also in the "followers" tab, click the "unfollow" button and approve the transaction.
11. Note that the buttons for this user in both tabs are now in the updating state (but nothing has changed otherwise).
12. Wait for the transaction to go through.
13. Note that this user is now removed from the "following" tab, and has a "follow" button in the "followers" tab.
14. Note that you can no longer see this person's posts in the "My Feed" section.
15. Reload the page to demonstrate these changes are persisted.

### DIFF
--- a/src/components/ConnectionsListProfiles.tsx
+++ b/src/components/ConnectionsListProfiles.tsx
@@ -1,11 +1,16 @@
 import React from "react";
 import UserAvatar from "./UserAvatar";
-import { Button } from "antd";
+import { Button, Spin } from "antd";
 import { Profile } from "../utilities/types";
 import { HexString } from "@dsnp/sdk/dist/types/types/Strings";
-import { useAppSelector } from "../redux/hooks";
+import { useAppDispatch, useAppSelector } from "../redux/hooks";
 import { createProfile } from "@dsnp/sdk/core/activityContent";
 import { AnnouncementType } from "@dsnp/sdk/core/announcements";
+import { followUser, unfollowUser } from "../services/content";
+import {
+  updateRelationshipStatus,
+  RelationshipStatus,
+} from "../redux/slices/graphSlice";
 
 enum ListStatus {
   CLOSED,
@@ -14,18 +19,32 @@ enum ListStatus {
 }
 
 interface ConnectionsListProfilesProps {
+  userId: string;
   listStatus: ListStatus;
-  following: Record<string, boolean>;
-  followers: Record<string, boolean>;
+  following: Record<string, RelationshipStatus>;
+  followers: Record<string, RelationshipStatus>;
 }
 
 const ConnectionsListProfiles = ({
+  userId,
   listStatus,
   following,
   followers,
 }: ConnectionsListProfilesProps): JSX.Element => {
-  const userRelationship = (userProfile: Profile) =>
-    following[userProfile.fromId] ? "Unfollow" : "Follow";
+  const dispatch = useAppDispatch();
+
+  const isFollowing = (userProfile: Profile): boolean =>
+    userProfile.fromId in following;
+
+  const isFollowingUpdating = (userProfile: Profile): boolean =>
+    following[userProfile.fromId] === RelationshipStatus.UPDATING;
+
+  const userRelationship = (profile: Profile): string =>
+    isFollowingUpdating(profile)
+      ? "updating"
+      : isFollowing(profile)
+      ? "Unfollow"
+      : "Follow";
 
   const profiles: Record<HexString, Profile> = useAppSelector(
     (state) => state.profiles?.profiles || {}
@@ -49,6 +68,21 @@ const ConnectionsListProfiles = ({
       ? Object.keys(following).map(profileForId)
       : [];
 
+  const changeGraphState = async (profile: Profile) => {
+    if (isFollowing(profile)) {
+      await unfollowUser(BigInt(userId), BigInt(profile.fromId));
+    } else {
+      await followUser(BigInt(userId), BigInt(profile.fromId));
+    }
+    dispatch(
+      updateRelationshipStatus({
+        follower: userId,
+        followee: profile.fromId,
+        status: RelationshipStatus.UPDATING,
+      })
+    );
+  };
+
   return (
     <>
       {connectionsList.map((userProfile) => (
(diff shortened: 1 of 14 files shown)


src/redux/slices/graphSlice.ts
```diff
@@ -2,9 +2,20 @@ import { DSNPUserId } from "@dsnp/sdk/dist/types/core/identifiers";
 import { createSlice, PayloadAction } from "@reduxjs/toolkit";
 import { GraphChange } from "../../utilities/types";
 
-interface graphState {
-  following: Record<DSNPUserId, Record<DSNPUserId, boolean>>;
-  followers: Record<DSNPUserId, Record<DSNPUserId, boolean>>;
+export enum RelationshipStatus {
+  ESTABlISHED,
+  UPDATING,
```
wilwade, line 7: Updating doesn't store the data about if the update is an update to follow or an update to unfollow. Should it?

src/redux/slices/graphSlice.ts
```diff
@@ -2,9 +2,20 @@ import { DSNPUserId } from "@dsnp/sdk/dist/types/core/identifiers";
...
+}
+
+export interface graphState {
+  following: Record<DSNPUserId, Record<DSNPUserId, RelationshipStatus>>;
+  followers: Record<DSNPUserId, Record<DSNPUserId, RelationshipStatus>>;
```
wilwade, line 12: Hmm... What is the followers RelationshipStatus mean?

Is it the relationship of the follower to the current user or the other way around? The follower to me will always be ESTABLISHED (as otherwise they are not a follower and will not be in the list). The other way around could have a relationship status of... well there isn't one, but `NONE` I guess.

src/redux/slices/graphSlice.ts
wilwade, line 7: Updating doesn't store the data about if the update is an update to follow or an update to unfollow. Do we want to know which way the update is moving?

src/components/ConnectionsListProfiles.tsx
```diff
@@ -50,6 +69,21 @@ const ConnectionsListProfiles = ({
...
+        follower: userId,
+        followee: profile.fromId,
+        status: RelationshipStatus.UPDATING,
+      })
+    );
+  };
+
   return (
     <>
       {connectionsList.map((userProfile) => (
```
wilwade, line 89: Not really part of this, but I found `userProfile` to be very confusing when trying to read the changes below. `connectionProfile` might be more clear in which profile we are referring?

src/components/ConnectionsListProfiles.tsx
```diff
@@ -15,18 +20,32 @@ enum ListStatus {
...
-    following[userProfile.fromId] ? "Unfollow" : "Follow";
+  const dispatch = useAppDispatch();
+
+  const isFollowing = (userProfile: Profile): boolean =>
+    userProfile.fromId in following;
+
+  const isFollowingUpdating = (userProfile: Profile): boolean =>
+    following[userProfile.fromId] === RelationshipStatus.UPDATING;
+
+  const userRelationship = (profile: Profile): string =>
```
wilwade, line 43: Same here as below. `userRelationship` is a bit unclear. `connectionRelationship` perhaps? Or just `relationshipStr` even?

src/components/ConnectionsListProfiles.tsx
```diff
@@ -15,18 +20,32 @@ enum ListStatus {
...
+  userId,
   listStatus,
   following,
   followers,
 }: ConnectionsListProfilesProps): JSX.Element => {
-  const userRelationship = (userProfile: Profile) =>
-    following[userProfile.fromId] ? "Unfollow" : "Follow";
+  const dispatch = useAppDispatch();
+
+  const isFollowing = (userProfile: Profile): boolean =>
```
wilwade, line 37: There are a few things below that harp on this, but `userProfile` makes me thing it is the logged in user or the user of the current profile, not the connection user which is what I think it is.

src/services/sdk.ts
```diff
@@ -51,6 +52,11 @@ export const getProfile = async (fromId: HexString): Promise<Profile> => {
   return profile;
 };
 
+const convertDSNPIdToURI = (id: DSNPUserId) =>
```
wilwade, line 55: Side note that this gets easier once sdk v3 is in.

src/components/ConnectionsListProfiles.tsx
```diff
@@ -68,17 +102,23 @@ const ConnectionsListProfiles = ({
...
-              }
-            >
-              &#10005;
-            </div>
+            {isFollowingUpdating(userProfile) ? (
+              <Spin></Spin>
+            ) : (
+              <div
+                className={
+                  userRelationship(userProfile) === "Follow"
```
wilwade, line 114: Isn't this just `isFollowing`?

wilwade: These nested function calls make me wonder if this should be a separate component and you can just map over that with all the calls being just params going into the functional component.

wilwade (dismissed):
I think it all works. The comments are mostly just me being confused by relationships and which direction the relationships is going.

I wonder if there is a better way to represent the graph data structure, but not sure there is an easy one in js that would be any more useful here.

src/redux/slices/graphSlice.ts
acruikshank: We don't currently care. We're either following someone (established), not following (indicated by absence), or somewhere in-between (updating). This is purely for display state, and updating just means the user has changed the relationship, but we haven't yet seen the change on chain. We display both transitions in the same way.

src/redux/slices/graphSlice.ts
acruikshank: The status is the status of the relationship itself.  Updating mostly means the relationship is in flux and the user doesn't get to make further changes until it's resolved one way or the other.

src/components/test/ConnectionsList.test.tsx
```diff
@@ -8,9 +8,9 @@ import { getPrefabProfile } from "../../test/testProfiles";
 const profile = getPrefabProfile(0);
 const graphs = getPreFabSocialGraph();
 const store = createMockStore({
-  user: { profile: profile },
-  profiles: { profiles: [] },
-  graphs: { graphs: graphs },
+  user: { id: profile.fromId },
```
shannonwells, line 11: ❤️

src/redux/slices/graphSlice.ts
```diff
@@ -35,20 +46,42 @@ export const graphSlice = createSlice({
             ...state.following,
             [follower]: {
               ...(state.following[follower] || {}),
-              [followee]: true,
+              [followee]: RelationshipStatus.ESTABlISHED,
```
shannonwells, line 49: I wish we could change this to "followed" but not for this PR

shannonwells (approved):
I just gave it a quick lookover.  I'm wondering if this has been tested against devnet or qanet, which are slower and might reveal potential race conditions or other bugs.  If not save it for acceptance testing

enddynayn (approved):

PR #83, opened 2021-09-02
split sdk package into dsnp and content and document

acruikshank:
Purpose
---------------
As the example client is meant to demonstrate how to use the DSNP sdk to build DSNP functionality, the code needs to be well documented and understandable. This is especially true in the services where most of the DSNP specific code lives. The `sdk` service package has become a little unruly with functions that wrap the DSNP sdk, store and retrieve content, and interact with the redux store.

Solution
---------------
This PR refactors the `sdk` package into a `dsnp` package that wraps DSNP SDK functionality for specific use within the example client, and a `content` service that mediates all interactions between DSNP, the store, and redux. It then adds documentation for all methods in both packages.

Change summary
---------------
* Rename `sdk` package to `dsnp`.
* Add `content` package.
* Move all functionality related to redux or the content store from `dsnp` to `content` (sometimes splitting the functions).
* Document all functions in `dsnp`
* Document all functions in `content`

Steps to Verify
----------------
1. This is a pure refactor, so e everything should behave exactly as it did before.


src/App.test.tsx
```diff
@@ -4,8 +4,10 @@ import * as wallet from "../src/services/wallets/wallet";
 
 import { mount } from "enzyme";
 import { componentWithStore, createMockStore } from "./test/testhelpers";
-import * as sdk from "./services/sdk";
+import * as dsnp from "./services/dsnp";
+import * as content from "./services/content";
 import * as hooks from "./redux/hooks";
+import { UnsubscribeFunction } from "@dsnp/sdk/dist/types/core/contracts/utilities";
```
wilwade, line 10:
```suggestion
import { UnsubscribeFunction } from "@dsnp/sdk/core/contracts/utilities";
```

src/services/content.ts
```diff
@@ -0,0 +1,344 @@
...
+
+/**
+ * storeActivityContent stores activity content on the static server
+ * @param content the content to store
+ * @returns hash of content (from which url can be derived)
+ */
+const storeActivityContent = async (
+  content: ActivityContentNote | ActivityContentProfile
+): Promise<string> => {
+  const hash = keccak256(core.activityContent.serialize(content));
```
wilwade, line 332: Likely not for this PR, but the sdk now has hash exposed and it adds the `0x` prefix (although not sure you want that here.)
```suggestion
  const hash = core.utilities.hash(core.activityContent.serialize(content));
```

src/services/content.ts
```diff
@@ -0,0 +1,344 @@
...
+  dispatch: Dispatch,
+  message: BroadcastAnnouncement,
+  blockNumber: number,
+  blockIndex: number,
+  batchIndex: number
+) => {
+  const decoder = new TextDecoder();
+
+  try {
+    const url = decoder.decode((message.url as any) as Uint8Array);
```
wilwade, line 259: Conflict and rebase warning with #81 which ever gets in first

src/services/content.ts
```diff
@@ -0,0 +1,344 @@
...
+// Content Package
+// This package manages content interaction between the DSNP network, storage, and the redux store.
+// It also provides functions to initiate the creation of new content.
+//
+
+//
+// Parameter types
+//
+
+type Dispatch = ThunkDispatch<any, Record<string, any>, AnyAction>;
```
wilwade, line 38: These anys worry me.

src/services/content.ts
```diff
@@ -0,0 +1,344 @@
...
+  const decoder = new TextDecoder();
+
+  dispatch(
+    upsertProfile({
+      ...profile,
+      fromId: decoder.decode((message.fromId as any) as Uint8Array),
+      blockNumber,
+      blockIndex,
+      batchIndex,
+    } as Profile)
```
wilwade, line 216: Why do we need the `as`?

src/services/content.ts
```diff
@@ -0,0 +1,344 @@
...
+  update: RegistryUpdateLogData
+) => {
+  dispatch(
+    upsertProfile({
+      ...createProfile(),
+      fromId: core.identifiers.convertDSNPUserURIToDSNPUserId(
+        update.dsnpUserURI
+      ),
+      handle: update.handle,
+    } as Profile)
```
wilwade, line 235: Why do we need the `as`?

src/services/dsnp.ts
```diff
@@ -0,0 +1,267 @@
...
+  SignedReplyAnnouncement,
+  GraphChangeAnnouncement,
+  SignedProfileAnnouncement,
+  SignedAnnouncement,
+} from "@dsnp/sdk/core/announcements";
+import { BatchPublicationLogData } from "@dsnp/sdk/core/contracts/subscription";
+import { WalletType } from "./wallets/wallet";
+import torusWallet from "./wallets/torus";
+import { DSNPUserId } from "@dsnp/sdk/dist/types/core/identifiers";
+import { UnsubscribeFunction } from "@dsnp/sdk/dist/types/core/contracts/utilities";
```
wilwade, line 19:
```suggestion
import { UnsubscribeFunction } from "@dsnp/sdk/core/contracts/utilities";
```

src/services/content.ts
```diff
@@ -0,0 +1,344 @@
...
+  GraphChangeAnnouncement,
+  DSNPGraphChangeType,
+} from "@dsnp/sdk/core/announcements";
+import { BatchPublicationLogData } from "@dsnp/sdk/core/contracts/subscription";
+import { upsertGraph } from "../redux/slices/graphSlice";
+import * as dsnp from "./dsnp";
+import { UnsubscribeFunction } from "@dsnp/sdk/dist/types/core/contracts/utilities";
+import { keccak256 } from "web3-utils";
+import { HexString } from "@dsnp/sdk/dist/types/types/Strings";
+import { DSNPUserId } from "@dsnp/sdk/dist/types/core/identifiers";
```
wilwade, line 26:
```suggestion
import { UnsubscribeFunction } from "@dsnp/sdk/core/contracts/utilities";
import { keccak256 } from "web3-utils";
import { HexString } from "@dsnp/sdk/types/Strings";
import { DSNPUserId } from "@dsnp/sdk/core/identifiers";
```

src/services/dsnp.ts
```diff
@@ -0,0 +1,267 @@
...
+ * @throwss if batch cannot be read.
+ */
+export const readBatchFile = async (
+  batchAnnouncement: BatchPublicationLogData,
+  rowHandler: AnnouncementRowHandler
+): Promise<void> => {
+  let batchIndex = 0;
+  const reader = await core.batch.openURL(
+    (batchAnnouncement.fileUrl.toString() as any) as URL
+  );
```
wilwade, line 156: Not yet, but once sdk v3.0.2 is out this can just be `batchAnnouncement.fileUrl`

src/services/dsnp.ts
```diff
@@ -0,0 +1,267 @@
...
+ * @returns true if object appears to be a graph change announcment
+ */
+export const isGraphChangeAnnouncement = (
+  obj: unknown
+): obj is GraphChangeAnnouncement => {
+  return (
+    (obj as Record<string, unknown>)["announcementType"] ===
+    AnnouncementType.GraphChange
+  );
+};
```
wilwade, line 174: This does a bit more testing, but it is the same thing otherwise.
```suggestion
export const isGraphChangeAnnouncement = core.announcements.validation.isGraphChangeAnnouncement;
```

src/services/dsnp.ts
```diff
@@ -0,0 +1,267 @@
...
+
+/**
+ * isBroadcastAnnouncment returns whether the provided object is likely a BroadcastAnnouncment
+ * @param obj any object
+ * @returns true if object appears to be a broadcast announcment
+ */
+export const isBroadcastAnnouncement = (
+  obj: unknown
+): obj is BroadcastAnnouncement => {
```
wilwade, line 183: There are separate validators for broadcast, reply and profile: https://libertydsnp.github.io/sdk-ts/modules/core_announcements_validation.html

src/services/fakesdk.ts
```diff
@@ -1,24 +0,0 @@
-import { HexString, Profile } from "../utilities/types";
```
wilwade, line 1: w00t deleted! 🎉

wilwade (dismissed):
A most of my notes are actually on the code that is just moving, not the actual new changes, so feel free to wrap those up as a separate PR if you prefer.

src/services/dsnp.ts
acruikshank: Yeah, the point of these is to segregate the content, not to validate it. There's been problems with validation in the past. Maybe we're past that now, but it would be nice if the SDK provided options for both.

wilwade: I think with v3 of the sdk we are past it, but if you just segregate, we shouldn't have the type narrowing or your later code that assumes the type can fail or be broken in odd ways.

src/services/dsnp.ts
```diff
@@ -0,0 +1,267 @@
...
+//
+
+/**
+ * setupProvider initializes the DSNP sdk with an ethereum provider.
+ * This function must be called with a Metamask or Torus WalletType before
+ * any other DSNP functionality is used (including subscriptions).
+ * @param walletType which wallet plugin we should use to retrieve the chain provider.
+ * @throws if walletType is unrecognized or wallet plugin cannot supply a provider.
+ */
+export const setupProvider = (walletType: WalletType): void => {
```
enddynayn, line 53: nice

src/services/dsnp.ts
```diff
@@ -0,0 +1,267 @@
...
+export const startSubscriptions = async (
+  batchHandler: BatchAnnouncementHandler,
+  registryHandler: RegistryUpdateHandler
+): Promise<UnsubscribeFunction> => {
+  let blockNumber: number;
+  let blockIndex = 0;
+
+  // subscribe to all announcements
+  const unsubscribeToBatchPublications = await core.contracts.subscription.subscribeToBatchPublications(
+    (announcement: BatchPublicationLogData) => {
```
enddynayn, line 118: does this help you to simply this? I think it was recently added.

https://github.com/LibertyDSNP/sdk-ts/blob/86fe0062c07aa54632f708ccf402223fd522a5eb/src/core/contracts/utilities.ts#L80

enddynayn (approved):
looking good

src/services/content.ts
acruikshank: This goes away with the SDK 3.0 changes (a PR stacked on this one). It's required, because the parsed values coming out of Parquet don't match the Profile type, so we have to coerce it.

src/services/content.ts
acruikshank: This is resolved in the SDK 3.0 PR. It's due to a mismatch in types, which wasn't a good idea.

src/services/content.ts
acruikshank: #81 has gone away. #86 supersedes it and is already rebased on this PR.

src/services/content.ts
acruikshank: This has to change more drastically because, while hash is now available, serialize is now inaccessible. At any rate, a problem for the SDK PR.

src/services/dsnp.ts
acruikshank: Nice! yes it does. There's another PR for the upgrade to 3.0 so I'll fix it there.

PR #85, opened 2021-09-02
add utility to build react upload host url

aneyzberg:
Purpose
---------------
We needed a small function to differentiate when we need a fully qualified url or just the information for the 
`REACT_APP_UPLOAD_HOST`. 

This is because we are serving the react app statically from the static server in our Dockerfile, and so the `REACT_APP_UPLOAD_HOST` is the same as the base url for the static sever. In certain places in the code we use `fetch` to resolve a url - which is able to resolve an empty `REACT_APP_UPLOAD_HOST` to the base url of the server. However in other places we are just building a URL and so need to find the fully qualified URL. 

Solution
----------------
Write a small utility function that will return a qualified url when needed. 

Change summary
---------------
* Add a small utility function that will return a qualified Url when needed. 
* Add a spec

### DIFF
--- a/src/services/sdk.ts
+++ b/src/services/sdk.ts
@@ -9,6 +9,7 @@ import { addFeedItem, clearFeedItems } from "../redux/slices/feedSlice";
 import { upsertProfile } from "../redux/slices/profileSlice";
 import { AnyAction, ThunkDispatch } from "@reduxjs/toolkit";
 import { Store } from "./Storage";
+import { buildBaseUploadHostUrl } from "../utilities/buildBaseUploadHostUrl";
 import {
   ActivityContentNote,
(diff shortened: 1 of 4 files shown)


src/services/sdk.ts
```diff
@@ -354,8 +354,10 @@ const storeActivityContent = async (
   content: ActivityContentNote | ActivityContentProfile
 ): Promise<string> => {
   const hash = keccak256(core.activityContent.serialize(content));
+  console.log("process env store activity", process.env);
```
enddynayn, line 357: i think you left this on accident

aneyzberg: thank uou. fixed.

acruikshank (approved):
Looks great

PR #86, opened 2021-09-03
upgrade to sdk 3

acruikshank:
Purpose
---------------
Significant breaking changes were made in SDK 3.0. This PR makes necessary changes to incorporate these changes.

Solution
---------------
The SDK now returning DSNPUserId as a BigInt drives most of the changes in this PR. BigInt is not serializable, so it can't be stored in Redux and it cannot be used as a map key. Our approach is to convert all id's to strings (as they were before) before storing them and occasionally converting them back when submitting content back to the SDK. 

One of the bigger changes is that we now create a completely new `Profile` type to be used within the example-client rather than union-ing SDK types and adding fields. This allows us to normalize some behavior between registry updates and profile upgrades. 

We also take advantage of some of the new utility functions the SDK provides. 

Change summary
---------------
* Upgrade to SDK 3.0
* Refactor to convert DSNPUserId to strings
* Refactor Profile type
* Fix tests

Steps to Verify
----------------
1. This is a pure refactor. No behavior should have changed.

### DIFF
--- a/src/components/NewPost.tsx
+++ b/src/components/NewPost.tsx
@@ -8,7 +8,6 @@ import { createNote } from "../services/Storage";
 import { sendPost } from "../services/content";
 import { createProfile } from "@dsnp/sdk/core/activityContent";
 import { FromTitle } from "./FromTitle";
-import { DSNPUserId } from "@dsnp/sdk/dist/types/core/identifiers";
 import { postLoading } from "../redux/slices/feedSlice";
 
 interface NewPostProps {
@@ -23,9 +22,7 @@ const NewPost = ({ onSuccess, onCancel }: NewPostProps): JSX.Element => {
   const [postMessage, setPostMessage] = React.useState<string>("");
   const [isValidPost, setIsValidPost] = React.useState<boolean>(false);
 
-  const userId: DSNPUserId | undefined = useAppSelector(
-    (state) => state.user.id
-  );
+  const userId: string | undefined = useAppSelector((state) => state.user.id);
 
   const profiles: Record<HexString, Profile> = useAppSelector(
     (state) => state.profiles?.profiles || {}
@@ -42,7 +39,7 @@ const NewPost = ({ onSuccess, onCancel }: NewPostProps): JSX.Element => {
   };
 
   const createPost = async () => {
-    if (!profile) return;
+    if (!userId || !profile) return;
     const newPostFeedItem: FeedItem = await createNote(
       postMessage,
       uriList,
--- a/src/components/NewPost.tsx
+++ b/src/components/NewPost.tsx
@@ -39,7 +39,7 @@ const NewPost = ({ onSuccess, onCancel }: NewPostProps): JSX.Element => {
   };
 
   const createPost = async () => {
-    if (!userId || !profile) return;
+    if (!userId) return;
     const newPostFeedItem: FeedItem = await createNote(
       postMessage,
       uriList,
--- a/src/components/RegistrationModal.tsx
+++ b/src/components/RegistrationModal.tsx
@@ -74,7 +74,7 @@ const RegistrationModal = ({
         walletAddress,
         formValues.handle
       );
-      onIdResolved(core.identifiers.convertDSNPUserURIToDSNPUserId(userURI));
+      onIdResolved(core.identifiers.convertToDSNPUserId(userURI).toString());
     } catch (error) {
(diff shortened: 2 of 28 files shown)


enddynayn (approved):

enddynayn:
fabulous!

src/components/RegistrationModal.tsx
```diff
@@ -74,7 +74,7 @@ const RegistrationModal = ({
         walletAddress,
         formValues.handle
       );
-      onIdResolved(core.identifiers.convertDSNPUserURIToDSNPUserId(userURI));
+      onIdResolved(core.identifiers.convertToDSNPUserId(userURI).toString());
```
shannonwells, line 77: This seems needlessly complicated since userURI is already a string.  We should have an SDK function that just strips the scheme prefix and returns the ID as a string.  Not for this PR obviously though.

src/services/content.ts
```diff
@@ -170,18 +170,17 @@ const dispatchFeedItem = (
...
   blockNumber: number
 ) => {
-  const decoder = new TextDecoder();
   dispatch(
     addFeedItem({
-      fromId: decoder.decode((message.fromId as any) as Uint8Array),
+      fromId: message.fromId.toString(),
       blockNumber: blockNumber,
-      hash: decoder.decode((message.contentHash as any) as Uint8Array),
+      hash: message.contentHash,
```
shannonwells, line 177: #relief

src/services/dsnp.ts
```diff
@@ -110,26 +115,35 @@ export const startSubscriptions = async (
...
   // subscribe to all announcements
+  let batchBlockNumber: number;
+  let batchBlockIndex = 0;
   const unsubscribeToBatchPublications = await core.contracts.subscription.subscribeToBatchPublications(
     (announcement: BatchPublicationLogData) => {
-      if (announcement.blockNumber !== blockNumber) {
-        blockNumber = announcement.blockNumber;
-        blockIndex = 0;
+      if (announcement.blockNumber !== batchBlockNumber) {
+        batchBlockNumber = announcement.blockNumber;
```
shannonwells, line 124: Non-blocking: We should consider abstracting this since the logic is repeated below.

shannonwells (approved):
LGTM only non-blocking comments

src/components/RegistrationModal.tsx
```diff
@@ -88,7 +88,7 @@ const RegistrationModal = ({
     registration: registry.Registration
   ): string | undefined =>
     profiles[
-      core.identifiers.convertDSNPUserURIToDSNPUserId(registration.dsnpUserURI)
+      core.identifiers.convertToDSNPUserId(registration.dsnpUserURI).toString()
```
claireolmstead, line 91: again here

aneyzberg (approved):
lgtm barring comments made by other folks

src/components/RegistrationModal.tsx
acruikshank: Totally agreed.

src/services/dsnp.ts
acruikshank: This logic actually goes away now that we're getting transactionIndex from the SDK.

src/utilities/buildBaseUploadHostUrl.ts
```diff
@@ -1,5 +1,5 @@
-export const buildBaseUploadHostUrl = (qualifiedUrl: boolean): string => {
-  return qualifiedUrl
-    ? `${window.location.host}/${process.env.REACT_APP_UPLOAD_HOST}`
-    : `${process.env.REACT_APP_UPLOAD_HOST}`;
+export const buildBaseUploadHostUrl = (): string => {
+  return process.env.REACT_APP_UPLOAD_HOST
+    ? process.env.REACT_APP_UPLOAD_HOST
```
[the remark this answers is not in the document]
acruikshank: good point.

src/components/NewPost.tsx
```diff
@@ -42,7 +39,7 @@ const NewPost = ({ onSuccess, onCancel }: NewPostProps): JSX.Element => {
   };
 
   const createPost = async () => {
-    if (!profile) return;
+    if (!userId || !profile) return;
```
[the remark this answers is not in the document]
acruikshank: Another good point. We need userId, but not profile. fixed.

PR #88, opened 2021-09-07
Add GraphChangeButton to Profile

acruikshank:
Purpose
---------------
We would like to be able to follow users whose content is posted in our general feed. We currently only have follow buttons in the connections list section, which means that we can only follow users who already follow us (and unfollow those users or the users we already follow). Now that we can view user's profiles by clicking on the header of a post, we can place a graph change button in the profile block to follow these users.

Solution
---------------
Add a GraphChangeButton to the profile page if it is not displaying the current user.

Change summary
---------------
* Add GraphChangeButton to the profile page.
* Fix the connection list so it displays the follows and followed by's of the other user (this was missed in earlier PRs)
* Refactor the styling so GraphChangeButton gets its own styles.

Steps to Verify
----------------
1. Start with a clean,  populated app.
2. In the general feed, scroll down the the first post by a user that is not the current user.
3. Click the header of the post to view the profile of this user on the right.
4. Note the new follow/unfollow button under the icon of the selected user in profile block to the right.
5. Keep clicking on post headers till you find a user that you are not currently following.
5. Click the button to follow this user, accept the signature requests, and wait for the transaction to complete.
6. Note that the button changes state.
7. Click the "back" button in the feed tab to show the current user profile.
8. Note that the followed user appears in the "following" tab and their posts appear in your feed.
9. Return to the new followed user's profile.
10. Click "unfollow" and complete transaction.
11. Confirm user is unfollowed in all the same places.

----------------

-Screen Shot 2021-09-07 at 12 36 38 PM


src/components/ConnectionsListProfiles.tsx
```diff
@@ -34,6 +34,10 @@ const ConnectionsListProfiles = ({
     (state) => state.profiles?.profiles || {}
   );
 
+  const userFollowing = useAppSelector(
```
claireolmstead, line 37: userImFollowing?  I feel like we can make this a little more descriptive

src/components/GraphChangeButton.tsx
```diff
@@ -54,7 +54,7 @@ const GraphChangeButton = ({
 
   return (
     <Button
-      className="ConnectionsListProfiles__button"
+      className="GraphChangeButton"
```
claireolmstead, line 57: I love this name change

src/components/Profile.tsx
```diff
@@ -6,10 +6,15 @@ import { useAppSelector } from "../redux/hooks";
 import * as types from "../utilities/types";
 import { saveProfile } from "../services/content";
 import { core } from "@dsnp/sdk";
+import GraphChangeButton from "./GraphChangeButton";
 
 const Profile = (): JSX.Element => {
   const userId: string | undefined = useAppSelector((state) => state.user.id);
 
+  const following = useAppSelector(
```
claireolmstead, line 14: for the sake of continuity with connectionsListProfiles, should we use `userImFollowing` or `userFollowing`?

src/components/ConnectionsListProfiles.tsx
acruikshank: This is a map of all users the current logged in user is following (as opposed to users the display user is following). So `userImFollowing1 is probably not what we want. I guess we could get really verbose. `usersCurrentUserIsFollowing`. Slightly shorter: `followedByCurrentUser`? I kind of like that.

src/components/Profile.tsx
acruikshank: This is actually, the other map of the users the displayed user is following. I guess `followedByDisplayedUser`? I'm not crazy about the passive voice. `displayUserFollowingProfiles`? I still think the passive voice is clearer.

claireolmstead (approved):

PR #91, opened 2021-09-08
changes made to storing batch announcements and note attchment typing

claireolmstead:
Purpose
---------------
[https://www.pivotaltracker.com/story/show/179376055](url)

Currently, on QA the loading of the feed fails because we are unsafely casting to the type of ActivityContentAttachment[] when somehow one of the pieces of data on QAnet has a null.

Solution
---------------
* We should be safely casting the content to the Activity Content type by validating the content and using a type narrowing is so that we can trust our types.

Change summary
---------------
- change the way that we are storying the type of batch announcements to be the correct type instead of storing them all as broadcasts.
- update types to stop using "as ..." and instead type check using "is"

Steps to Verify
----------------
1. make sure that null data coming through QA net does not throw an error anymore.

### DIFF
--- a/src/components/Post.tsx
+++ b/src/components/Post.tsx
@@ -6,10 +6,10 @@ import PostMedia from "./PostMedia";
 import RelativeTime from "./RelativeTime";
 import ReplyBlock from "./ReplyBlock";
 import PostHashDropdown from "./PostHashDropdown";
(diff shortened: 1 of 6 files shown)


acruikshank (approved):
Looks good

PR #99, opened 2021-09-14
Lazy load activity content

acruikshank:
Purpose
---------------
Currently the example client attempts to load all existing content on the DSNP network at startup. Obviously, this is not going to scale. This PR reworks the app pretty significantly (but mostly behind the scenes), to load content only when we need to show it. Specifically it introduces and "Infinite scroll" feature to the feed to display older posts only after scrolling down to them. It then lazy loads all the activity content (posts, replies and profiles) so they are only loaded when viewed.

Solution
---------------
The first thing to note is that this PR introduces `react-query` to fetch activity content. This is somewhat controversial because this project already uses redux, so react-query is in some ways a second store. The motivation is that we want to load activity content only when it is viewed, but it may be viewed in multiple places simultaneously and will be re-rendered often. We could try to manage this by saving the content in redux, but we would need to worry about cache invalidation and we'd end up adding a lot of logic that does exactly what react-query does. It's better to think of it like the UI always goes out to the internet for activity content, only we put a good caching layer in front of it.

Another thing to note is that the example client had been doing several things wrong when it comes to announcements. This PR was forced to fix those.

Change summary
---------------
* Fix populate script to generate announcements of the correct type.
* Introduce a react-query QueryProvider at the top of the app.
* Modify FeedItem to contain only announcement information.
* Generally replace `Profile` (ProfileAnnouncement | ActivityContentProfile) with `User` (ProfileAnnouncement).
* Modify subscription and redux logic to only store announcements without first loading the activity content.
* Introduce separate `replies` structure in feed slice, so replies can be stored separately from posts and be more easily retrieved by inReplyTo field.
* Introduce infinite scroll component in PostList.
* Load activityContent of reply in Reply component.
* Load activityContent of post in the Post component.
* Modify `UserAvatar` to take a `User` and load the profile with the icon.
* Extract FromName component and a `userIdentifier` function to standardize logic of naming users that may not have a profile or handle.
* Fix announcement creation in sendPost, sendReply and saveProfile.
* Miscellaneous cleanup. 
* Fix tests.

Steps to Verify
----------------
1. Set up a clean populated app.
2. Note that everthing still works.
3. In the network tab, you might notice feed items loading as you scroll down the page. You probably won't though because the app tends to make all feed items visible briefly as they load from the subscription, which completely primes the react-query cache.


package.json
```diff
@@ -32,7 +32,9 @@
     "react": "17.0.1",
     "react-app-polyfill": "^2.0.0",
     "react-dom": "17.0.1",
+    "react-infinite-scroll-component": "^6.1.0",
     "react-player": "^2.9.0",
+    "react-query": "^3.23.2",
```
enddynayn, line 37: cool

scripts/populate.mjs
```diff
@@ -460,11 +460,23 @@ const storeAnnouncement = async (content, accountId, signer) => {
 
   // create and store announcement of content to batch
   const announcement = core.announcements.sign(
-    core.announcements.createBroadcast(
-      accountId,
-      `${process.env.REACT_APP_UPLOAD_HOST}/${hash}.json`,
-      hash
-    ),
+    type === core.announcements.AnnouncementType.Broadcast
```
enddynayn, line 463: this would be nice if we wrapped this its own function

scripts/populate.mjs
```diff
@@ -489,6 +501,35 @@ const storeAvatar = async (handle) => {
   );
 };
 
+const publishAnnouncements = async (announcements) => {
+  const hash = announcements.reduce(
```
enddynayn, line 505: nice

src/components/PostList.tsx
```diff
@@ -19,59 +21,73 @@ interface PostListProps {
...
 
-  const feed: FeedItem[] = sortFeed(initialFeed);
+  useEffect(() => {
+    const feedTypeFilters = {
+      [FeedTypes.MY_FEED]: (p: FeedItem) =>
+        p.fromId === userId ||
+        (myGraph && myGraph[p.fromId]?.status === RelationshipStatus.FOLLOWING),
+      [FeedTypes.MY_POSTS]: (p: FeedItem) => userId === p?.fromId,
+      [FeedTypes.DISPLAY_ID_POSTS]: (p: FeedItem) => displayId === p?.fromId,
+      [FeedTypes.DISCOVER]: (_p: FeedItem) => true,
```
enddynayn, line 58: maybe add this one in here are unknown ((_p) => true)

enddynayn (approved):
great!

scripts/populate.mjs
acruikshank: good call. done.

src/components/PostList.tsx
acruikshank: I'd like to do that, but there isn't a FeedType.UNKNOWN enum value. The unexpected case is handled on line 61 with the `|| ((_p) => true)`

PR #101, opened 2021-09-15
authorize announcement messages before dispatching

acruikshank:
Purpose
---------------
The example client should not accept improperly signed announcements. This PR rectifies that.
I needed to check that EC was sending valid signatures, and adding this authorization was the easiest way. Now that it's there I don't see any reason to get rid of it.

Solution
---------------
Add check to ignore announcements if a) they're not SignedAnnocements b) their signature does not resolve to a designated authority with announce privileges on chain. 

Change summary
---------------
* Add checks described above.

Steps to Verify
----------------
1. Should work the same as it used to.
2. You can modify the code to produce a bad signature and confirm an error appears in the console and the post does not appear in the feed.

### DIFF
--- a/src/services/content.ts
+++ b/src/services/content.ts
@@ -259,7 +259,7 @@ const handleRegistryUpdate = (dispatch: Dispatch) => (
 const handleBatchAnnouncement = (dispatch: Dispatch) => (
   batchAnnouncement: BatchPublicationLogData
 ) => {
-  dsnp.readBatchFile(batchAnnouncement, (announcementRow, batchIndex) => {
+  dsnp.readBatchFile(batchAnnouncement, async (announcementRow, batchIndex) => {
     try {
       const announcement = announcementRow as unknown;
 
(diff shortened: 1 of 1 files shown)


src/services/content.ts
```diff
@@ -272,6 +298,8 @@ const handleBatchAnnouncement = (dispatch: Dispatch) => (
           })
         );
       } else if (isBroadcastAnnouncement(announcement)) {
+        console.log("broadcast annoucement url", announcement.url);
+        console.log("broadcast annoucement hash", announcement.contentHash);
```
wilwade, line 302: These look like extra 🪵

src/services/content.ts
```diff
@@ -260,6 +262,30 @@ const handleBatchAnnouncement = (dispatch: Dispatch) => (
...
+          announcement,
+          BigInt(announcement.fromId),
+          Permission.ANNOUNCE
+        )
+      ) {
+        console.log(
+          "batched announcment has unauthorized signature",
+          announcement
+        );
+        return;
```
wilwade, line 286: Do we want to show invalid stuff, but flag it? I guess this is a good first piece either way.

wilwade (changes requested):
Just some extra logs in the isBroadcastAnnouncement check

src/services/content.ts
acruikshank: I don't think so. At best it would be vector to pollute the feed. At worst it would leave doubt about the authenticity of counterfeit content. There's also no real way to do it for profiles or graph changes without allowing malicious behavior.

src/services/content.ts
acruikshank: good catch. I forgot to push a commit.

wilwade (approved):

PR #106, opened 2021-09-16
restore lib type checking

acruikshank:
Purpose
---------------
Our typescript configuration had skipLibCheck set to true, which permitted some invalid type usage throughout the app. This PR turns that check back on and resolves all the issues that the check catches.

Solution
---------------
Stop the skipLibCheck and fix a few problems that appeared once the check is on.

Change summary
---------------
* skipLibCheck set to false.
* Create a custom `FeedItem` type with the values we actually use, rather aliasing `BroadcastAnnouncement` (which isn't close to how we're using it.
* Rename `Reply` to `ReplyItem` and define it like `FeedItem`.
* Fix everyplace we type something DSNPUserId when we're actually using a string.
* Rename `FeedItem.uri` to `FeedItem.url` (finally).
* Fix errors coming from the react-query queries.

Steps to Verify
----------------
1. Test normally, there should be no user-facing changes.


wilwade (approved):
One non-blocking extra import

PR #109, opened 2021-09-20
feed reset after logout on network change

claireolmstead:
Purpose
---------------
[https://www.pivotaltracker.com/story/show/179380420](https://www.pivotaltracker.com/story/show/179380420)
Restart
Feature: Clear feed and logout on network change. logout on account change,

Solution
---------------
- clear feed and logout in the chainChanged listener

Change summary
---------------
* clear feed and logout in the chainChanged listener
* change naming from log in to connect

Steps to Verify
----------------
1. test account change
2. test network change
3. look for expected response

### DIFF
--- a/src/components/Login.tsx
+++ b/src/components/Login.tsx
@@ -65,7 +65,7 @@ const Login = (): JSX.Element => {
     }
   };
 
-  const login = async (selectedType: wallet.WalletType) => {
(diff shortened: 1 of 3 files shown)


shannonwells (approved):
This clears the feed but it doesn't clear graph and profiles when changing networks, which I think we should also do, but I think it's fine to do this in another PR.

PR #110, opened 2021-09-20
misc styling changes

claireolmstead:
Purpose
---------------
feature
[https://www.pivotaltracker.com/story/show/179648866](https://www.pivotaltracker.com/story/show/179648866)
As a user, we want the the the UI to have consecutive design. As a developer, we want naming to be the same between classes and files.

Solution
---------------
- make changes to scss
- change class and file names where necessary
- delete mocked blank post and blank reply and replace with undefined

Change summary
---------------
* name changes
* change border radius' to either 5 or 10px
* change styling on the connections list buttons
* make login spinner green
* UserAvatar should be passed undefined for user instead of blank user in BlankPost and BlankReply

Steps to Verify
----------------
1. Check that it looks good and functions as expected

----------------
- Screenshot


shannonwells (approved):
LGTM

PR #111, opened 2021-09-20
style from title and metadata of posts and replies

claireolmstead:
Purpose
---------------
feature
[https://www.pivotaltracker.com/story/show/179649000](https://www.pivotaltracker.com/story/show/179649000)
create prettier UI and fix timestamp bug

Solution
---------------
- change styling for reply and post from title
- add the ability to click on users' profile from replies
- convert timestamp from unix bigint to ISO when you call addFeedItem
- move the timestamp to the top right corner

Change summary
---------------
* described in solution

Steps to Verify
----------------
1. Run code
2. create new post
3. see that it looks good and there are no errors

----------------
- Screenshot


src/services/content.ts
```diff
@@ -300,6 +300,12 @@ const handleBatchAnnouncement = (dispatch: Dispatch) => (
           })
         );
       } else if (isBroadcastAnnouncement(announcement)) {
+        let publishedDate;
+        if (announcement.createdAt > (2 ^ 53) - 1) {
```
shannonwells, line 304:
```suggestion
        if (announcement.createdAt > Number.MAX_SAFE_INTEGER) {
```
This is an actual value: Number.MAX_SAFE_INTEGER and should be used instead of being calculated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

shannonwells (approved):
For expediency I'm marking a conditional approve assuming the one change to use Number.MAX_SAFE_INTEGER, and I made it in suggestion format so you could possibly just accept it from the PR.

PR #113, opened 2021-09-20
fix timestamp bug, misc styling, don't show choose registration unles…

claireolmstead:
Purpose
---------------
[https://www.pivotaltracker.com/story/show/179649749](url)
Feature

Solution
---------------
- should not show switch registrations unless there are 2+
- don't render handle and social address in the input blocks anymore since they are not inputs
- remove decrement in the monthNum of relative time
- remove the dots at the end of the date
- make small user avatar smaller

Change summary
---------------
* look to solution description

Steps to Verify
----------------
1. run code
2. makes sure it looks good and doesn't throw errors

----------------
- Screenshot

### DIFF
--- a/src/components/EditRegistrationAccordion.tsx
+++ b/src/components/EditRegistrationAccordion.tsx
@@ -47,7 +47,7 @@ const EditRegistrationAccordion = ({
           onIdResolved={onIdResolved}
         />
       </Panel>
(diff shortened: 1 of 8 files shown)


shannonwells (approved):
This looks good.  THANK YOU for splitting it up! Makes me so happy, so much easier to evaluate.

PR #115, opened 2021-09-21
Feat/friendly error messages

acruikshank:
Purpose
---------------
The general goal of this PR is to translate errors messages that come from the network and rarely relate to what the user is trying to accomplish into helpful instructions. In the process of working out what these errors were, I found and fixed a few bugs.

Solution
---------------
Introduce a "friendlyErrors" function that knows how to find the relevant text of an error message and translates common errors into friendlier messages.

Change summary
---------------
* Introduce `friendlyErrors`.
* Add popup error handling for logins driving by wallet changes and profile changes.
* Fix issue where registration modal would put app in bad state when user clicks outside it.
* Fix issue where handle updates would fail if user was auto-logged in with the session.
* Updated `antd` (possibly not necessary but still a good idea).
* Replaced custom `UnsubscribeError` with type from SDK now that it is exposed.
* Fixed bug related to above change.

Steps to Verify
----------------
1. Log out. Set MetaMask to an account with multiple registrations. Log in and choose MetaMask. Click outside the registraiton modal. All modals should be closed and the login button should not be in the loading state. You should be able to log in normally after that.
2. Log in. Reload the page. You should still be logged in  through the session. Click your handle at the top right, Choose "Create New Handle". Create a new handle. It should not fail with an obscure error about DNS names (the from address should be defined).
3. Set MetaMask to point to an Ethereum mainnet address and try to log in. You should get an understandable message about what's wrong and what you need to do to fix it.
4. Try to change your handle to a handle that already exists. You should get a straightforward error message saying what's wrong and what you need to do to fix it.
5. Try to change your handle or your profile name, but refuse to sign the message. You should get an understandable "Transaction Canceled" error.

### DIFF
--- a/src/components/Login.tsx
+++ b/src/components/Login.tsx
@@ -49,12 +49,16 @@ const Login = (): JSX.Element => {
     setRegistrationPopoverVisible(false);
   };
 
-  // clear state if something clears the wallet type (e.g. an error connecting to the provider)
   useEffect(() => {
     if (currentWalletType === wallet.WalletType.NONE) {
-      startLoading(false);
-      setPopoverVisible(false);
-      setRegistrationVisible(false);
+      // close modals if something clears the wallet type (e.g. an error connecting to the provider)
+      closeModals();
+    } else {
+      // Ensure wallet address is set for all  components this component owns.
+      (async () => {
+        const waddr = await wallet.wallet(currentWalletType).getAddress();
+        setWalletAddress(waddr);
+      })();
     }
   }, [currentWalletType]);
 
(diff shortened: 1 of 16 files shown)


src/components/Login.tsx
```diff
@@ -76,15 +89,19 @@ const Login = ({ isPrimary, loginWalletOptions }: LoginProps): JSX.Element => {
       const waddr = await wallet.wallet(selectedType).login();
       await loginWithWalletAddress(waddr, selectedType);
     } catch (error) {
+      console.log("Login error", error);
```
shannonwells, line 92: If we're trapping it but it's an error, maybe this would be better logged as console.warn.

src/components/RegistrationModal.tsx
```diff
@@ -46,6 +48,13 @@ const RegistrationModal = ({
     }
   }
 
+  const handleVisbileChange = (visible: boolean) => {
```
shannonwells, line 51:
```suggestion
  const handleVisibleChange = (visible: boolean) => {
```
typo

shannonwells (approved):
Very friendly!  2 nits and then there's the conflict between your & Claire's changes. The typo should be fixed though.

PR #116, opened 2021-09-21
edit profile picture modal

claireolmstead:
Purpose
---------------
Feature
As a user, I should be able to edit my profile picture
[https://www.pivotaltracker.com/story/show/179324830](https://www.pivotaltracker.com/story/show/179324830)

Solution
---------------
- onClick of the edit button in the profile component you should be able to edit your profile picture with a new URL.

Change summary
---------------
* Create EditAvatarModal that gives the user a UI to upload their new profile photo and returns the ActivityContentImageLink value.
* if edited, save the new profile icon in the Profile component
* create temp icon uri state that displays that new icon until the real iconURL is completely loaded
* styling for the new UI
* test for the new UI
* if you cancel it clears inputs/states

Steps to Verify
----------------
1. click edit
2. click edit on profile photo
3. put in valid url
4. try invalid url
5. try add in both cases
6. try cancel from modal
7. try cancel after new image is uploaded
8. click save after new image is uploaded

---------------

https://user-images.githubusercontent.com/43625033/134262675-c6192cf0-9ba6-4b2f-82f3-f08d03c8e94f.mov

### DIFF
--- a/src/components/EditAvatarModal.tsx
+++ b/src/components/EditAvatarModal.tsx
@@ -1,0 +1,95 @@
+import React from "react";
+import UserAvatar from "./UserAvatar";
+import { useAppDispatch, useAppSelector } from "../redux/hooks";
+import * as types from "../utilities/types";
+import { Alert, Button, Input, Modal } from "antd";
+import { useState } from "react";
+import { CameraOutlined } from "@ant-design/icons";
+import { ActivityContentImageLink } from "@dsnp/sdk/core/activityContent";
+import { core } from "@dsnp/sdk";
+import path from "path";
+import { setTempIconUri } from "../redux/slices/userSlice";
+
+interface EditAvatarModalProps {
+  setNewIcon: (icon: ActivityContentImageLink[] | undefined) => void;
+}
+
+const EditAvatarModal = ({ setNewIcon }: EditAvatarModalProps): JSX.Element => {
+  const dispatch = useAppDispatch();
+
+  const userId: string | undefined = useAppSelector((state) => state.user.id);
+  const profiles: Record<types.HexString, types.User> = useAppSelector(
+    (state) => state.profiles?.profiles || {}
+  );
+  const user: types.User | undefined = userId ? profiles[userId] : undefined;
+
+  const [isModalVisible, setIsModalVisible] = useState<boolean>(false);
+  const [iconUri, setIconUri] = React.useState<string>("");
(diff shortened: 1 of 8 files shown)


src/components/Profile.tsx
```diff
@@ -52,24 +63,29 @@ const Profile = (): JSX.Element => {
 
   const saveEditProfile = async () => {
```
claireolmstead, line 64: we don't need to check that because we already do through didEditProfile

src/components/test/EditAvatarModal.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    );
+    component.find(".EditAvatarModal__edit").simulate("click");
+    component.find("input").simulate("change", {
+      target: { value: "https://placekitten.com/300/300" },
+    });
+    component
+      .find(".NewPostImageUpload__urlInputBtn")
+      .first()
+      .simulate("click");
+    expect(component.find("input").props().value).toEqual(
```
claireolmstead, line 47: @shannonwells and I were trying to get this to simulate clicking on add and seeing the avatar change and we had a hard time with it.  If you have any suggestions, let me know.

src/components/EditAvatarModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [iconUri, setIconUri] = React.useState<string>("");
+  const [urlErrorMsg, setUrlErrorMsg] = React.useState<string | null>(null);
+  const hasValidPostURI = iconUri && String(iconUri).match(/.+\..+\//);
+
+  const handleNewUri = () => {
+    if (hasValidPostURI) {
+      const extension = path.extname(iconUri).replace(".", "");
+      const imageLink = core.activityContent.createImageLink(
+        iconUri,
+        `image/${extension}`,
```
acruikshank, line 36: This is a little dicey. There's no guarantee the URL will have the correct extension (or especially that the URL will end with it). Also not clear that the image will be one of the [supported types](https://spec.dsnp.org/ActivityContent/Associated/Attachments#supported-image-mime-types). That level of validation will be tough though.

For this PR you should lower case the extension and then check that it is one of the supported types. If it isn't, it's up to you whether you should reject the request or make it "jpeg" and hope for the best.

acruikshank (approved):
Looks good. I'd do something less risky when deriving the type from the extension.

PR #124, opened 2021-10-04
Bug/follow unfollow graph state update/179326255

claireolmstead:
Purpose
---------------
As a user, I want to see the correct follower/following list based on my graph.

Solution
---------------
Set and filter the connectionsList in the parent component, `ConnectionsList` instead of in the child component.  We were not initially filtering out the users that did not have the correct status, therefor giving us an inaccurate count in cases such as if you are following then unfollow a user.

Change summary
---------------
* made connectionsList a function that is passed the list that we want to filter.
* calls connectionsList(followers/following).length to get followers/following count
* calls connectionsList(currentRelation) to get open connections list and sends that as a prop to `ConnectionsListProfiles`;

Steps to Verify
----------------
1. Follow a user
2. unfollow that user
3. see that the followers/following count is correct
4. follow that user again
5. see that followers/following count is correct

### DIFF
--- a/src/components/ConnectionsList.tsx
+++ b/src/components/ConnectionsList.tsx
@@ -10,9 +10,9 @@ import { User } from "../utilities/types";
 import { AnnouncementType } from "@dsnp/sdk/core/announcements";
 
 enum ListStatus {
-  CLOSED,
-  FOLLOWERS,
-  FOLLOWING,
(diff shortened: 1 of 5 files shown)


enddynayn (approved):
👍 fabuloso.

PR #125, opened 2021-10-07
Feature/display video thumbnail/179326255

claireolmstead:
Purpose
---------------
As a user, I would like to be able to see a thumbnail for the video that I am attempting to include in a new post rather than the default icon, which looks like there is an error with the link.

Solution
---------------
- If there is a thumbnail in the metadata, show that thumbnail.
- else, display a video icon so it at least looks valid

Change summary
---------------
* remove console.logs that got in from former PR
* created `getThumbnail` function that checks if youtube, vimeo, or image, and returns the thumbnail URL or null;
* if thumbnail is null, display alt thumbnail image
* add scss for alt thumbnail

Steps to Verify
----------------
- works for Youtube, Vimeo, mp4, and any other accepted video formats
- if you upload a link that is valid but does not provide a thumbnail, such as a Soundcloud link, see the catch-all UI.

----------------
Thumbnails in order of links:
- soundcloud
- youtube
- image
- vimeo

Screen Shot 2021-10-07 at 12 07 26 PM

### DIFF
--- a/src/components/NewPostImageUpload.tsx
+++ b/src/components/NewPostImageUpload.tsx
@@ -43,6 +43,16 @@ const NewPostImageUpload = ({
       <div className="NewPostImageUpload__cover">
         {uriList &&
           uriList.map((image, index) => {
+            const isYoutubeVideo = image.match(
+              /^https?\:\/\/(?:www\.youtube(?:\-nocookie)?\.com\/|m\.youtube\.com\/|youtube\.com\/)?(?:ytscreeningroom\?vi?=|youtu\.be\/|vi?\/|user\/.+\/u\/\w{1,2}\/|embed\/|watch\?(?:.*\&)?vi?=|\&vi?=|\?(?:.*\&)?vi?=)([^#\&\?\n\/<>"']*)/i
+            );
+            if (isYoutubeVideo) {
+              image =
+                "https://img.youtube.com/vi/" +
+                isYoutubeVideo[1] +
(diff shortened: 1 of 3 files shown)


src/components/NewPostImageUpload.tsx
```diff
@@ -38,14 +44,45 @@ const NewPostImageUpload = ({
     setUriList(deletedImageArray);
     onNewPostImageUpload(deletedImageArray);
   };
+
+  const getThumbnail = (uri: string): string | undefined => {
+    const isYoutubeVideo = uri.match(
+      /^https?\/\/(?:www\.youtube(?:nocookie)?\.com\/|m\.youtube\.com\/|youtube\.com\/)?(?:ytscreeningroom\?vi?=|youtu\.be\/|vi?\/|user\/.+\/u\/\w{1,2}\/|embed\/|watch\?(?:.*)?vi?=|vi?=|\?(?:.*)?vi?=)([^#\n/<>"']*)/i
```
shannonwells, line 50: Are we using this (or a subset of this) anywhere else?  Might be good to extract it (and the other two patterns) into something reusable.

shannonwells (approved):
Code all looks good to me.

src/components/NewPostImageUpload.tsx
claireolmstead: we are not. we're checking but not filtering in the same way because we just care if it's a video in react player.  in this case, we need to know more specific details.

PR #128, opened 2021-10-11
Added function to check for whitespace on display name and trim

waylandli:
Purpose
---------------
https://www.pivotaltracker.com/n/projects/2436354/stories/179382064
Users were able to put display names of " ", "                  ", "          test        ".
The proposed changes removes the ability to have a whitespace as a display name as well as trimming the names to not have space on either side.

Solution
---------------
Before creating the profile to be saved as a change, the name goes through a new function to check for empty whitespace as well as trim the sides of the name for any excess whitespace.

Change summary
---------------
* Moved the creation of profile to inside the try/catch block
* Made function to check for any unnecessary whitespace that is called before profile creation
* Added a testcase to check that the correct display name is shown after saving
* Removed section of friendly error as it no longer applies to every type of error

Steps to Verify
----------------
1. Try to make a name with only spaces and verify the error popup shows up
2. Make a name with spaces on the sides and check that they are trimmed correctly


enddynayn:
looking good. try adding a little test case.

waylandli:
Alright I'll try making some tests and the necessary changes to get past the checks

PR #129, opened 2021-10-12
updated regex for youtube videos

claireolmstead:
Purpose
---------------
bug
[https://www.pivotaltracker.com/story/show/179914342](https://www.pivotaltracker.com/story/show/179914342)
maybe a character got deleted in the regex or something but isYoutubeVideo is returning null when it should not be.

Solution
---------------
- make the regex correct so we can see the thumbnail.

Change summary
---------------
* Changed regex to something more simple

Steps to Verify
----------------
1. create new post with youtube video and see thumbnail.

----------------
working!


src/components/NewPostImageUpload.tsx
```diff
@@ -45,12 +45,14 @@ const NewPostImageUpload = ({
 
   const getThumbnail = (uri: string): string | undefined => {
     const isYoutubeVideo = uri.match(
-      /^https?\/\/(?:www\.youtube(?:nocookie)?\.com\/|m\.youtube\.com\/|youtube\.com\/)?(?:ytscreeningroom\?vi?=|youtu\.be\/|vi?\/|user\/.+\/u\/\w{1,2}\/|embed\/|watch\?(?:.*)?vi?=|vi?=|\?(?:.*)?vi?=)([^#\n/<>"']*)/i
+      /\/\/((?:www|m)\.)?((?:youtube\.com|youtu.be))(\/(?:[\w-]+\?v=|embed\/|v\/)?)([\w]+)(\S+)?$/i
```
shannonwells, line 48: I was about to 💣  this because I wrote a quick unit test and it failed both of the youtube link examples until you updated it. I know it's fixed now, but there should be a unit test for this and other pattern matches (like the Vimeo one below). If it's abstracted into a function in a library (which could be useful), it could be tested.  This will also catch accidental overwrites/typos from people's editors and such later.

src/utilities/helpers.ts
```diff
@@ -0,0 +1,15 @@
+export const getThumbnail = (url: string): string | undefined => {
```
shannonwells, line 1: ❤️

PR #133, opened 2021-10-18
timestamp + tests

claireclark1:
Purpose
---------------
bug
[https://www.pivotaltracker.com/story/show/179957621](https://www.pivotaltracker.com/story/show/179957621)
As a user, I want to see when posts were published.

Solution
---------------
check out content.ts and find bug in saving the timestamp to the store.
make sure it appears correctly in the feed
is there a better way to test?

Change summary
---------------
* found bug and fixed if statement to return the correct timestamp.
* 

Steps to Verify
----------------
1. create post
2. check that timestamp is correct
3. see posts coming in from other users and on other networks and see that timestamp is correct

----------------
Screen Shot 2021-10-18 at 9 24 42 AM

### DIFF
--- a/src/components/Profile.tsx
+++ b/src/components/Profile.tsx
@@ -66,11 +66,14 @@ const Profile = (): JSX.Element => {
   const saveEditProfile = async () => {
     setIsEditing(!isEditing);
     if (userId === undefined) return;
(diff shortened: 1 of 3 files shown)


shannonwells (commented):
... "+ tests?"  Weren't you going to add a test?

PR #135, opened 2021-10-25
Removed torus wallet connection

waylandli:
Purpose
---------------
https://www.pivotaltracker.com/n/projects/2436354/stories/180031983

We wanted to delete functionality for torus connection as we want to develop the client using metamask for now.

Change summary
---------------
- Looked for torus files and deleted them
- Deleted mentions of options that allow torus
- deleted test cases of torus functionality
- You can connect the wallet through one click rather than clicking through the modal
- Removed log in modal
- Simplified some wallet code


claireolmstead (approved):
🥇

PR #137, opened 2021-11-02
Feature/embedded url/177294663

claireolmstead:
Purpose
---------------
Feature

[https://www.pivotaltracker.com/story/show/177294663](https://www.pivotaltracker.com/story/show/177294663)

As a user, when I include a link in the text of a post or reply, it should be clickable.  OnClick, the link is opened in a new tab.

Solution
---------------
When reading the content for posts and replies we want to look for valid URL, replace them with a link tag.

styling:
if a valid URL, it should turn the URL to dark green.

Change summary
---------------
* add a find valid link helper function
* use when displaying content for a post or reply
* add testing

Steps to Verify
----------------
1. Put a link in a caption
2. put a link in a reply
3. click on it
4. see that it takes you to a new page

----------------
- it was too much of a pain to auto find the URL when you're typing and change the styling, so we just kept that black and only inserted the colored link in the saved post and replies

Screen Shot 2021-11-02 at 12 23 47 PM


src/components/FromTitle.tsx
```diff
@@ -15,9 +15,9 @@ export const FromTitle = ({
   isHoveringProfile,
   isReply,
 }: FromTitleProps): JSX.Element => {
-  const atHandle = userInfo.handle && "@" + userInfo.handle;
+  const atHandle = userInfo?.handle && "@" + userInfo?.handle;
```
claireolmstead, line 18: issues with testing so had to add the "?"s in

src/components/test/Feed.test.tsx
```diff
@@ -139,4 +139,15 @@ describe("Feed", () => {
...
     });
   });
+
+  it("displays link correctly in post", () => {
+    const component = mount(componentWithStore(Feed, store));
+    expect(
+      component.find(".Post__caption").first().prop("dangerouslySetInnerHTML")
+        ?.__html
+    ).toEqual(
+      'test post <a class="messageLink" href=https://www.unfinishedlabs.io/ target="_blank">https://www.unfinishedlabs.io/</a>'
```
claireolmstead, line 149: linter does not like this.  tried `` as well but prettier changed them to ''. any suggestions?

wilwade: @claireolmstead Let's update the .eslintrc.json file rules to this (which will fix it and allow `'I "like" this rule better'`):

OLD: `"quotes": ["error", "double"],`
NEW: `"quotes": ["error", "double", { "avoidEscape": true, "allowTemplateLiterals": true }],`

src/utilities/helpers.ts
```diff
@@ -10,3 +10,10 @@ export const getThumbnailUrl = (url: string): string => {
 
   return url;
 };
+
+export const formatMessage = (message: string): string => {
```
wilwade, line 14: I think we might want a utility package that does this work for us. Urls can be much more complex. Here are two that might be worth checking on: https://www.npmjs.com/package/react-anchorme or http://markdown-it.github.io/linkify-it/doc/

src/components/FromTitle.tsx
wilwade: That would imply that our prop type is wrong. userInfo should exist, but handle might not. I wonder if we can fix the tests to take these optionals back out.

wilwade (commented):
Few comments.

wilwade (approved):
🚂 LTGM!

PR #138, opened 2021-11-03
fix docker deploy to update latest version when needed

waylandli:
Problem
---------------
[Finishes 179608496](https://www.pivotaltracker.com/n/projects/2436354/stories/179608496)

Currently our docker latest tag is not automatically updated when a new tag version of example client is pushed.

Solution
---------------
Changed docker workflow to update latest version when a new version of example client is pushed.

Change summary
---------------
* Updated the docker-release yaml file
* Checks if newest version is greater than the last latest version before updating

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -8,7 +8,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-
+        with:
+          fetch-depth: '0'
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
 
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-release.yml
```diff
@@ -20,10 +21,52 @@ jobs:
...
           echo ${GITHUB_REF##*/}
-          echo "TAG=${GITHUB_REF##*/}" >> $GITHUB_ENV
+          echo "NEW_TAG_VERSION=${GITHUB_REF##*/}" >> $GITHUB_ENV
+
+      - name: Find commit for latest tag
+        run:  |
+          COMMIT=$(git rev-list -n 1 docker/latest)
+          echo "COMMIT_FOR_LATEST_TAG=${COMMIT}" >> $GITHUB_ENV
+
+      - name: Setting latest tag version
```
aneyzberg, line 35: Find tag version associated with latest tag

.github/workflows/docker-release.yml
```diff
@@ -20,10 +21,52 @@ jobs:
...
+        run: |
+          LATEST_TAG_VERSION=$(git tag --list "docker/v*" --points-at ${{env.COMMIT_FOR_LATEST_TAG}})
+          echo "LATEST_TAG_VERSION=${LATEST_TAG_VERSION}" >> $GITHUB_ENV
+
+      - name: Checking Tag Versions
```
aneyzberg, line 40: Check Tag Versions

.github/workflows/docker-release.yml
```diff
@@ -20,10 +21,52 @@ jobs:
...
+        uses: actions/github-script@v5
+        id: set-result
+        env:
+          LATEST_TAG_VERSION: ${{env.LATEST_TAG_VERSION}}
+          NEW_TAG_VERSION: ${{env.NEW_TAG_VERSION}}
+        with:
+          script: |
+            const { LATEST_TAG_VERSION , NEW_TAG_VERSION } = process.env
+            console.log("Old tag version: ", LATEST_TAG_VERSION)
```
aneyzberg, line 49: Delete this console.log

.github/workflows/docker-release.yml
```diff
@@ -20,10 +21,52 @@ jobs:
...
+            console.log("New tag version: ", NEW_TAG_VERSION)
```
aneyzberg, line 50: delete this console.log

.github/workflows/docker-release.yml
```diff
@@ -20,10 +21,51 @@ jobs:
...
+            const latestTagVersion = LATEST_TAG_VERSION.split('docker/v')
+            const newTagVersion = NEW_TAG_VERSION.split('v')
+
+            const latestTagVersionArray = latestTagVersion[1].split('.')
+            const newTagVersionArray = newTagVersion[1].split('.')
+
+            const latestTagVersionNum = latestTagVersionArray.join("")
+            const newTagVersionNum = newTagVersionArray.join("")
+
+            if (newTagVersionNum > latestTagVersionNum) {
```
wilwade, line 63: Hmm... This will fail in cases of a double digit version number. `1.10.0` should be < `2.0.0`, but it isn't. I think the array entries must be tested independently for semantic versioning.

.github/workflows/docker-release.yml
```diff
@@ -32,7 +74,17 @@ jobs:
           push: true
           file: Dockerfile
           tags: |
-            dsnp/example-client:${{ env.TAG }}
+            dsnp/example-client:${{ env.NEW_TAG_VERSION }}
+
+      - name: Build and push dsnp/example-client:latest
+        if: steps.set-result.outputs.result == 'true'
+        id: docker_build_example-client-latest
+        uses: docker/build-push-action@v2
```
wilwade, line 82: Does this trigger a rebuild?

.github/workflows/docker-release.yml
```diff
@@ -20,10 +21,51 @@ jobs:
...
+            
+            if (!LATEST_TAG_VERSION) {
+              return "true"
+            }
+
+            const latestTagVersion = LATEST_TAG_VERSION.split('docker/v')
+            const newTagVersion = NEW_TAG_VERSION.split('v')
+
+            const latestTagVersionArray = latestTagVersion[1].split('.')
+            const newTagVersionArray = newTagVersion[1].split('.')
```
wilwade, line 58: We might want to do an early return of true if the length of either these two arrays is < 2.

wilwade (changes requested):
Few questions. Also while checking something in the review I stumbled upon https://github.com/docker/metadata-action which might do all this work for us...

.github/workflows/docker-release.yml
aneyzberg: can you elaborate on what you meant here?

.github/workflows/docker-release.yml
waylandli: We added a conditional to both build blocks to prevent rebuilding

wilwade (approved):

Issue #151, opened 2022-02-24
Replace node-sass with sass


wilwade:
node-sass is deprecated and does not work on newer M1 macs.

- https://sass-lang.com/blog/libsass-is-deprecated

Issue #150, opened 2021-11-30
Update to Node 16


wilwade:
# Summary

Now that Node 16 is LTS, we should update to use Node 16.

- [ ] .node-version
- [ ] .tool-versions
- [ ] GitHub Actions
- [ ] package-lock.json file (npm default changed with v16)

PR #152, opened 2022-03-08
Chore/replace node sass/151

mehtaishita:
Purpose
---------------
Replacing and updated LibSass version

Related GitHub Issues
---------------
https://github.com/LibertyDSNP/example-client/issues/151

Solution
---------------
Remove node-sass dependency and replace it with the newest version of Sass.

@wilwade 

Change summary
---------------
* Remove node-sass from package.json
* Replace it with sass version 1.49.9
* There is a type error in the implementation of dispatch(startSubscriptions). Decided to cast startSubscriptions as 'any' temporarily and added a comment
* Also needed to add missing babel packages in webpack config

Steps to Verify
----------------
1. Reinstall node_modules
2. Compile and smoke test

----------------


src/App.tsx
```diff
@@ -36,7 +36,8 @@ const App = (): JSX.Element => {
       try {
         await wallet.wallet(walletType).reload();
         await setupProvider(walletType);
-        unsubscribeFunction = await dispatch(startSubscriptions);
+        // TODO: correct generic type for startSubscriptions overload - and remove Any
+        unsubscribeFunction = await dispatch(startSubscriptions as any);
```
raebot, line 40: ❌ Do we need to include this change and TODO in this PR? It doesn't seem related to the `node-sass` > `sass` switch.

raebot (changes requested):
👍🏻 I pulled the branch, installed the new dependency, and ran tests. All worked for me. I left one recommendation for a small change, otherwise looks good. Ping me again and I can do a quick review.

src/App.tsx
wilwade: It's not, but on @mehtaishita box it was not compiling without it. After struggling with it for a while, we decided to punt on the fix.

raebot: Ah ok 👍🏻

raebot (approved):

src/App.tsx
wilwade: @raebot It was any build. I was unable to reproduce on my machine as well, but @mehtaishita has the cleanest machine, so I trust that computer more for issues.

I think the #150 might help solve the issue or show it clearly on other machines.

PR #153, opened 2022-03-10
Chore/update node 16/150

mehtaishita:
Purpose
---------------
Updating Nodejs from 14 to 16

Related GitHub Issues
---------------
#150

Solution
---------------
I updated the proper config and tool versions from node engine from 14.xx.xx to 16.14.0

Change summary
---------------
* Changed the asdf install to node 16
* Updated tool-version, GitHub actions YAML, package json

Steps to Verify
----------------
1. run: asdf install node 16
2. run: asdf which node (ensure the install is node 16.14.0
3. compile and run example-client


shannonwells (approved):
Thanks for this badly needed update.

wilwade (changes requested):
I missed a spot we have Node 14. The Dockerfile: https://github.com/LibertyDSNP/example-client/blob/main/Dockerfile#L1

